PR #3, opened 2015-10-23
Tags refactoring

ilyapuchka:
I noticed that at first tags were plain strings but then they changed to generics. In this pr I revert that change and here are my thoughts about that:

1. In my experience of working with DI (particularly in Obj-C with Typhoon) I've never needed named registrations. In most cases the same could be achieved by creating different containers for different environments or using runtime arguments.

2. Adding generic constraint on tag type adds type constraint on the whole container. I.e. it will be hard to add child-parent relationship between containers, or combine them in collection without introducing base protocol (what I think is useless at this point)

3. Adding generic tags limits containers flexibility and functionality. If you set it once you have to use only this type of tag inside container, so you have to move components that need other tags to other containers. It can lead to enormous number of containers. For example it make sense to group instances from network layer in one container, but that instances can have different responsibilities and it will be hard to mark them with tag of one type in context of one container, so you will end up with no tags at all or using strings or ints (ints are much less expressive than strings in that case) or you will create a lot of small containers.

4. At the end what our code can gain from using generic tags? Strings are rather expressive already. If you want safety you can always use string constants. Generics make containers code more complicated.

It would be interesting to know your motivation for moving to generic tags.

### DIFF
--- a/Example/DipSampleApp/DependencyContainers.swift
+++ b/Example/DipSampleApp/DependencyContainers.swift
@@ -27,8 +27,8 @@ let wsDependencies = DependencyContainer() { dip in
 }
 
 
-let NamedStarshipProviderTag    = Tag.String("NamedStarshipProvider")
-let PlistPersonProviderTag      = Tag.String("PlistPersonProvider")
+let NamedStarshipProviderTag: Tag    = "NamedStarshipProvider"
+let PlistPersonProviderTag: Tag      = "PlistPersonProvider"
 
 // MARK: Dependency Container for Providers
 let providerDependencies = DependencyContainer() { dip in
--- a/Example/DipSampleApp/DependencyContainers.swift
+++ b/Example/DipSampleApp/DependencyContainers.swift
@@ -27,8 +27,8 @@ let wsDependencies = DependencyContainer() { dip in
 }
 
 
-let NamedStarshipProviderTag    = "NamedStarshipProvider"
-let PlistPersonProviderTag      = "PlistPersonProvider"
+let NamedStarshipProviderTag    = Tag.String("NamedStarshipProvider")
+let PlistPersonProviderTag      = Tag.String("PlistPersonProvider")
 
 // MARK: Dependency Container for Providers
 let providerDependencies = DependencyContainer() { dip in
--- a/Example/DipSampleApp/DependencyContainers.swift
+++ b/Example/DipSampleApp/DependencyContainers.swift
@@ -19,7 +19,7 @@ private let FAKE_STARSHIPS = false
 
 
 // MARK: Dependency Container for WebServices & NetworkLayer
-let wsDependencies = DependencyContainer<WebService>() { dip in
+let wsDependencies = DependencyContainer() { dip in
     
     // Register the NetworkLayer, same for everyone here (but we have the ability to register a different one for a specific WebService if we wanted to)
     dip.register(instance: URLSessionNetworkLayer(baseURL: "http://swapi.co/api/")! as NetworkLayer)
@@ -27,15 +27,17 @@ let wsDependencies = DependencyContainer<WebService>() { dip in
 }
 
 
+let NamedStarshipProviderTag    = "NamedStarshipProvider"
+let PlistPersonProviderTag      = "PlistPersonProvider"
 
 // MARK: Dependency Container for Providers
-let providerDependencies = DependencyContainer<Int>() { dip in
+let providerDependencies = DependencyContainer() { dip in
     
     if FAKE_PERSONS {
         
         // 1) Register the PersonProviderAPI singleton, one generic and one specific for a specific personID
         dip.register(instance: DummyPilotProvider() as PersonProviderAPI)
-        dip.register(0, instance: PlistPersonProvider(plist: "mainPilot") as PersonProviderAPI)
+        dip.register(PlistPersonProviderTag, instance: PlistPersonProvider(plist: "mainPilot") as PersonProviderAPI)
         
     } else {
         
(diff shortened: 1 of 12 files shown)


AliSoftware:
Hi @ilyapuchka 

My motivations were simply that some people asked why the type was constrained to a `String` when it could be useful to make it anything as long as it's `Equatable`. Especially people wanted to use it for `enums` and `Int`.

But your rationale make sense, especially the fact that `DependencyContainer` introduce a type constraint that could make it hard to use. So I'm inclined to agree about improving that.

I think we could imagine an in-between solution there: the tag could be of a type `Tag`, which could be a dedicated class (or struct? or enum?) that would be `StringLiteralConvertible` and `IntergerLiteralConvertible`. This way, for people wanting/needing to use tags to differentiate instances, they could either use a string literal or an integer literal transparently, instead of being bound to `String`
They could even create an `extension` on this `Tag` class if they want to support more types for the tag.

In the end I don't use that tag feature much myself, so I'm not entirely sure how people use it, but it could still be interesting to allow them more flexibility… while still avoid adding constraints to the `DependencyContainer` like I did with the generic constraint.

Would be interested by your thoughts

---

PS: btw, I never used Typhoon (nor done real practical DI before) so I'd be very interested in your feedback of using `Dip` in everyday projects!

ilyapuchka:
Yes, I think `StringLiteralConvertible` will also work fine here. But again it will require either to keep constraint on container or to add generic constraint to each method that accepts tag as parameter.
And I don't see any need to separately support `IntergerLiteralConvertible`, it will just add additional methods with not much profit. I would prefer clean code more than generic tags.

AliSoftware:
I get you and feel the same but the integer tag was requested at some point. I've started a thing where the tag is an `enum` with associated value `String` or `Int` so people can choose what suit them, even if I personally rarely use those tags at all.

ilyapuchka:
@AliSoftware any progress on that?

AliSoftware:
Hi @ilyapuchka 

Sorry, didn't forget you, and still have your PR in mind, I just haven't have much time to get to it yet.

What do you say you add some commits to make the tag not just `String`, but `enum { case str(String), case int(Int) }`? That way the `DependencyContainer` won't be generic anymore, but still allowing some flexibility.
That's actually what I wanted to do on your PR before merging it, but got busy elsewhere in the meantime.

ilyapuchka:
@AliSoftware ok, will do =)

Example/DipSampleApp/ViewControllers/StarshipListViewController.swift
```diff
@@ -12,7 +12,9 @@ class StarshipListViewController : UITableViewController, FetchableTrait {
     var objects: [Starship]?
     var batchRequestID = 0
     
-    private func provider(tag:Int?) -> StarshipProviderAPI { return providerDependencies.resolve(tag) }
+    private func provider(tag:Int?) -> StarshipProviderAPI {
+        return providerDependencies.resolve( tag != nil ? .Int(tag!) : nil)
```
AliSoftware: Could use `tag.flatMap{.Int($0)}` there instead

README.md
```diff
@@ -84,21 +84,16 @@ let dip = DependencyContainer<String> { dip in
...
+* The tags can be StringLiteralType or IntegerLiteralType. That said you can use plain strings or integers as tags.
 
-For example, you can use an `Int`, a `String`… or even an `enum`, like this:
 
 ```swift
-enum WebService {
-    case PersonWS
-    case StarshipWS
-}
+let PersonWSTag: Tag = "PersonWS"
```
AliSoftware: What about using `enum WebService : String`, and then use the `.rawValue` for the tag? This would make nice "scoped constants" instead of keeping them wild at the global namespace level with those top-level `let`?

Example/DipSampleApp/DependencyContainers.swift
```diff
@@ -27,15 +27,17 @@ let wsDependencies = DependencyContainer<WebService>() { dip in
...
 // MARK: Dependency Container for Providers
-let providerDependencies = DependencyContainer<Int>() { dip in
+let providerDependencies = DependencyContainer() { dip in
     
     if FAKE_PERSONS {
         
         // 1) Register the PersonProviderAPI singleton, one generic and one specific for a specific personID
         dip.register(instance: DummyPilotProvider() as PersonProviderAPI)
-        dip.register(0, instance: PlistPersonProvider(plist: "mainPilot") as PersonProviderAPI)
+        dip.register(PlistPersonProviderTag, instance: PlistPersonProvider(plist: "mainPilot") as PersonProviderAPI)
```
AliSoftware: By changing that, you associated that `PersonProviderAPI` with a `String` tag instead of an `Int` tag. This means that the example project probably won't work as expected anymore, since when resolving, you kept [that](https://github.com/AliSoftware/Dip/pull/3/files#diff-ca934cf6a8bc3ae3d5d8e7afbcd16082R20) logic:

```swift
let provider = providerDependencies.resolve(.Int(personID)) as PersonProviderAPI
```

Probably better rollback that line to `dip.register(0, …)` then?

Example/DipSampleApp/DependencyContainers.swift
```diff
@@ -48,7 +50,7 @@ let providerDependencies = DependencyContainer<Int>() { dip in
 
         // 2) Register the StarshipProviderAPI factories, one generic and one specific for a specific starshipID
         dip.register() { HardCodedStarshipProvider() as StarshipProviderAPI }
-        dip.register(0) { DummyStarshipProvider(pilotName: "Main Pilot") as StarshipProviderAPI }
+        dip.register(NamedStarshipProviderTag) { DummyStarshipProvider(pilotName: "Main Pilot") as StarshipProviderAPI }
```
AliSoftware: Same comment as for the `dip.register` or the `PersonProviderAPI`

ilyapuchka:
@AliSoftware applied your comments

AliSoftware:
Looks good from here, will check back at home tomorrow.

Could you add a `CHANGELOG` entry for that change, crediting yourself? Thx

PR #5, opened 2015-11-01
Added support to register components with runtime arguments

ilyapuchka:
To be able to pass runtime arguments to `resolve()` we need to create set of register/resolve functions with different number of arguments. Then keys used to access factories should also take in account number of arguments used at registration.

### DIFF
--- a/Sources/Dip.swift
+++ b/Sources/Dip.swift
@@ -16,184 +16,278 @@ import Foundation
  */
 public class DependencyContainer {
   
-  /**
-   Use a tag in case you need to register multiple instances or factories
-   with the same protocol, to differentiate them. Tags can be either String
-   or Int, to your convenience.
-   */
-  public enum Tag: Equatable {
-    case String(StringLiteralType)
-    case Int(IntegerLiteralType)
-  }
+    /**
+     Use a tag in case you need to register multiple instances or factories
+     with the same protocol, to differentiate them. Tags can be either String
+     or Int, to your convenience.
+     */
+    public enum Tag: Equatable {
+        case String(StringLiteralType)
+        case Int(IntegerLiteralType)
+    }
 
-  /**
-   *  Internal representation of a key to associate protocols & tags to an instance factory
-   */
-  private struct LookupKey : Hashable, Equatable, CustomDebugStringConvertible {
-    var protocolType: Any.Type
-    var associatedTag: Tag?
+    private var dependencies = [DefinitionKey : Definition]()
+    private var lock: OSSpinLock = OS_SPINLOCK_INIT
     
-    var hashValue: Int {
-      return "\(protocolType)-\(associatedTag)".hashValue
+    /**
+     Designated initializer for a DependencyContainer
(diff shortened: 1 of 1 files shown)


AliSoftware:
Nice addition! :+1:

Just some random notes: There are two things I currently don't like about`Dip` so far:

1. It lacks Unit Tests of its own — the example project has Unit Tests to demonstrate how to use `Dip` in the context of App vs Tests, but the lib doesn't have tests on its own — so it's hard to check if something broke. Currently Travis only checks that the example project still compile.

2. The sample project is probably too complex for a sample. It started simple, but began to grow quite quickly — first in order to make it more like a real-world project, but now I think the sample doesn't concentrate enough on Dip and on how to demonstrate basic usage of the lib.

Those are obviously my own fault and are mainly due to lack of time to spend on Dip lately, but given those:

1. We can't really be sure that your PR works as expected, It'd be about time we add Unit Tests to `Dip` for that first
2. It'd probably be hard for people to see how to use your new addition, because the existing Sample is probably too complex already to add your additions into it and make it even harder to understand for a sample.
3. So this PR could actually be the occasion to create a different, simpler example, that concentrate better on `Dip` on its usage, and avoid a complex project with lots of concepts unrelated to `Dip` (like I feel my current sample did wrong)

If you had time to do one of these (probably the 2nd sample dedicated to the features you added via that PR at least) that would be really awesome!

AliSoftware:
PS: We may have to check and think about the various usages of this, but maybe your addition would make the `Tag` useless, as it could be replaced by a runtime parameter in many if not all cases?

I haven't thought this thru, but if that's the case, that could probably make the code even better and simpler, right?

ilyapuchka:
@AliSoftware I think tags should stay as optional parameter cause you can have cases when you need to register different implementations of the same protocol. If you use runtime argument for that you will need some kind of factory, that will return different implementations depending on tag value. Registering with tags makes more sense in this case. More than that components will be defined all in one place (when you register them), not in different factories.

Actually I though the same about sample project, which is a bit hard to follow, and lack of unit tests. But to add them we first need to add framework target for Dip itself, 'cause now it's build by Cocoapods. Also it will enable use of Carthage. Will try to invest some time in that next week.

AliSoftware:
> But to add them we first need to add framework target for Dip itself, 'cause now it's build by Cocoapods. Also it will enable use of Carthage.

Indeed. You could check out how I did that in my other repo `OHHTTPStubs` if you want some inspiration. Basically, I did multiple distinct projects: one that builds the framework, and that can be used by Carthage, and contains the Unit Tests for the lib; and other projects are just Sample projects that use the lib as a Development Pod (`pod 'OHHTTPStubs', :path => '../OHHTTPStubs'`).

AliSoftware:
_PS: I sent you an email (hope you don't mind) about Push Access; I hope I got your email right_

ilyapuchka:
I will close this PR now and will create new one after adding unit tests and some reorganizing of project.

PR #10, opened 2015-11-08
Playground

ilyapuchka:
This PR adds playground that demonstrates Dip's current features. I broke it in few pages for better experience.

@AliSoftware Please fill free to correct it and add what ever you think will be good to illustrate in this playground.

Also I'm not sure if there is a way to make Dip module accessible from playground's supplementary code, so I had to import it in every page and do some small setup at the top of some pages.


AliSoftware:
> Also I'm not sure if there is a way to make Dip module accessible from playground's supplementary code, so I had to import it in every page and do some small setup at the top of some pages.

I don't think that's possible. It seems that we indeed have to `import Dip` in every page, [according to Apple's documentation](https://developer.apple.com/library/ios/recipes/Playground_Help/Chapters/ImportFramework.html).

AliSoftware:
I reviewed the playground pages, that's a nice work, great job :+1: 

I think we could add a last page about "Shared Instances", explaining why they are bad design and how Dependency Container can help replacing them. Given what's explained in the previous pages that could seem to be implicit and restating some stuff already explained especially in the first page, but I thing closing with a typical example of what people do wrong and how to migrate away would be a nice conclusion.

Probably some page showing one or two typical dummy classes (fake ViewControllers maybe) using a `WebService.sharedInstance` internally (no DI, hard coupling), then right below the same classes implemented using Dip's `DependencyContainer` so the user can compare. Then ends the page by explaining that it also allows easier testing as we can register a `DummyService` in the Unit Tests, as you briefly explained in your intro.

I may try to add that some time (but given how busy I am with OHHTTPStubs, SwiftGen and my blog, you might probably beat me to it)

ilyapuchka:
Hey, @AliSoftware, thanks for review! I've added two more pages as you suggested - for tests and singletons.

AliSoftware:
Just fixed some typos. LGTM!

PR #11, opened 2015-11-11
Circular dependencies

ilyapuchka:
This PR adds support for circular dependencies. That feature and using `resolve` inside factory block requires recursion, but lock around `resolve` method causes deadlock. I removed locking completely and propose to managed thread safety on client side.

### DIFF
--- a/Dip/Dip/Definition.swift
+++ b/Dip/Dip/Definition.swift
@@ -81,12 +81,14 @@ public struct DefinitionOf<T, F>: Definition {
    ```
    
    */
-  public mutating func resolveDependencies(container: DependencyContainer, tag: DependencyContainer.Tag? = nil, block: (DependencyContainer, T) -> ()) {
+  public func resolveDependencies(container: DependencyContainer, tag: DependencyContainer.Tag? = nil, block: (DependencyContainer, T) -> ()) -> DefinitionOf<T, F> {
     guard resolveDependenciesBlock == nil else {
       fatalError("You can not change resolveDependencies block after it was set.")
     }
-    resolveDependenciesBlock = block
-    container.register(tag: tag, definition: self)
+    var newDefinition = self
+    newDefinition.resolveDependenciesBlock = block
+    container.register(tag: tag, definition: newDefinition)
+    return newDefinition
   }
   
   let factory: F
--- a/Dip/Dip/Definition.swift
+++ b/Dip/Dip/Definition.swift
@@ -81,23 +81,25 @@ public struct DefinitionOf<T, F>: Definition {
    ```
    
    */
-  public func resolveDependencies(container: DependencyContainer, tag: DependencyContainer.Tag? = nil, block: (DependencyContainer, T) -> ()) -> DefinitionOf<T, F> {
+  public func resolveDependencies(container: DependencyContainer, block: (DependencyContainer, T) -> ()) -> DefinitionOf<T, F> {
     guard resolveDependenciesBlock == nil else {
       fatalError("You can not change resolveDependencies block after it was set.")
     }
     var newDefinition = self
     newDefinition.resolveDependenciesBlock = block
-    container.register(tag: tag, definition: newDefinition)
+    container.register(newDefinition)
     return newDefinition
   }
   
   let factory: F
   var scope: ComponentScope
   var resolveDependenciesBlock: ((DependencyContainer, T) -> ())?
+  let tag: DependencyContainer.Tag?
   
-  init(factory: F, scope: ComponentScope) {
+  init(factory: F, scope: ComponentScope, tag: DependencyContainer.Tag?) {
     self.factory = factory
     self.scope = scope
+    self.tag = tag
   }
   
   ///Will be stored only if scope is `Singleton`
@@ -111,7 +113,7 @@ public struct DefinitionOf<T, F>: Definition {
   mutating func resolvedInstance(container: DependencyContainer, tag: DependencyContainer.Tag? = nil, instance: T) {
     guard scope == .Singleton else { return }
     _resolvedInstance = instance
-    container.register(tag: tag, definition: self)
+    container.register(self)
   }
   
   private var _resolvedInstance: T?
--- a/Dip/Dip/Definition.swift
+++ b/Dip/Dip/Definition.swift
@@ -57,17 +57,7 @@ public enum ComponentScope {
 }
 
 ///Definition of type T describes how instances of this type should be created when they are resolved by container.
-public final class DefinitionOf<T>: Definition {
-  
-  /**
-   Changes scope of the component.
-   
-   - parameter scope: new scope value. New definitions have `Prototype` scope
-  */
-  public func inScope(scope: ComponentScope) -> DefinitionOf<T> {
-    self.scope = scope
-    return self
-  }
+public struct DefinitionOf<T, F>: Definition {
   
   /**
    Sets the block that will be used to resolve dependencies of the component. 
@@ -84,23 +74,26 @@ public final class DefinitionOf<T>: Definition {
    ```swift
    container.register { [unowned container] ClientImp(service: container.resolve() as Service) as Client }
 
-   container.register { ServiceImp() as Service }
-    .resolveDependencies { container, service in
+   var definition = container.register { ServiceImp() as Service }
+   definition.resolveDependencies { container, service in
       service.delegate = container.resolve() as Client
    }
    ```
    
    */
-  public func resolveDependencies(block: (DependencyContainer, T) -> ()) -> DefinitionOf<T> {
-    self.resolveDependenciesBlock = block
-    return self
+  public mutating func resolveDependencies(container: DependencyContainer, tag: DependencyContainer.Tag? = nil, block: (DependencyContainer, T) -> ()) {
+    guard resolveDependenciesBlock == nil else {
+      fatalError("You can not change resolveDependencies block after it was set.")
+    }
+    resolveDependenciesBlock = block
+    container.register(tag: tag, definition: self)
   }
   
-  let factory: Any
+  let factory: F
   var scope: ComponentScope
   var resolveDependenciesBlock: ((DependencyContainer, T) -> ())?
   
-  init(factory: Any, scope: ComponentScope) {
+  init(factory: F, scope: ComponentScope) {
     self.factory = factory
     self.scope = scope
   }
@@ -111,10 +104,12 @@ public final class DefinitionOf<T>: Definition {
       guard scope == .Singleton else { return nil }
       return _resolvedInstance
     }
-    set {
-      guard scope == .Singleton else { return }
-      _resolvedInstance = newValue
-    }
+  }
+  
+  mutating func resolvedInstance(container: DependencyContainer, tag: DependencyContainer.Tag? = nil, instance: T) {
+    guard scope == .Singleton else { return }
+    _resolvedInstance = instance
+    container.register(tag: tag, definition: self)
   }
   
   private var _resolvedInstance: T?
--- a/Dip/Dip/Definition.swift
+++ b/Dip/Dip/Definition.swift
@@ -50,20 +50,62 @@ func ==(lhs: DefinitionKey, rhs: DefinitionKey) -> Bool {
 public enum ComponentScope {
   /// Indicates that a new instance of the component will be created each time it's resolved.
   case Prototype
+  /// Indicates that instances will be reused during resolve but will be discurded when topmost `resolve` method returns.
+  case ObjectGraph
   /// Indicates that resolved component should be retained by container and always reused.
   case Singleton
 }
 
 ///Definition of type T describes how instances of this type should be created when they are resolved by container.
 public final class DefinitionOf<T>: Definition {
-  let factory: Any
-  let scope: ComponentScope
   
-  init(factory: Any, scope: ComponentScope = .Prototype) {
+  /**
+   Changes scope of the component.
+   
+   - parameter scope: new scope value. New definitions have `Prototype` scope
+  */
+  public func inScope(scope: ComponentScope) -> DefinitionOf<T> {
+    self.scope = scope
+    return self
+  }
+  
+  /**
+   Sets the block that will be used to resolve dependencies of the component. 
+   This block will be called before `resolve` returns.
+   
+   - parameter block: block to use to resolve dependencies
+   
+   - note:  
+   If you have circular dependencies at least one of them should use this block
+   to resolve it's dependencies. Otherwise code enter infinite loop.
+   
+   **Example**
+   
+   ```swift
+   container.register { [unowned container] ClientImp(service: container.resolve() as Service) as Client }
+
+   container.register { ServiceImp() as Service }
+    .resolveDependencies { container, service in
+      service.delegate = container.resolve() as Client
+   }
+   ```
+   
+   */
+  public func resolveDependencies(block: (DependencyContainer, T) -> ()) -> DefinitionOf<T> {
+    self.resolveDependenciesBlock = block
+    return self
+  }
+  
+  let factory: Any
+  var scope: ComponentScope
+  var resolveDependenciesBlock: ((DependencyContainer, T) -> ())?
+  
+  init(factory: Any, scope: ComponentScope) {
     self.factory = factory
     self.scope = scope
   }
   
+  ///Will be stored only if scope is `Singleton`
   var resolvedInstance: T? {
     get {
       guard scope == .Singleton else { return nil }
--- a/Dip/Dip/Dip.swift
+++ b/Dip/Dip/Dip.swift
@@ -68,13 +68,38 @@ public class DependencyContainer {
     definitions.removeAll()
   }
   
+  /**
+   Removes previously registered definition from container.
+   
+   - parameter tag: tag used to register definition
+   - parameter definition: definition to remove
+  */
+  public func remove<T, F>(tag tag: Tag? = nil, definition: DefinitionOf<T, F>) {
+    let key = DefinitionKey(protocolType: T.self, factoryType: F.self, associatedTag: tag)
+    definitions[key] = nil
+  }
+  
+  /**
+   Registers new definiton in container and associate it with provided tag.
+   Will override already registered definition for the same type and factory associated with the same tag.
+   
+   - parameter tag: The arbitrary tag to associate definition with
+   - parameter definition: definition to register in container
+  */
+  public func register<T, F>(tag tag: Tag? = nil, definition: DefinitionOf<T, F>) {
+    let key = DefinitionKey(protocolType: T.self, factoryType: F.self, associatedTag: tag)
+    definitions[key] = definition
+  }
+  
   // MARK: Register dependencies
   
   /**
   Register a Void->T factory associated with optional tag.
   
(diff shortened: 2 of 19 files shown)


ilyapuchka:
@AliSoftware hey, could you take a look when you have time?

AliSoftware:
Missed that, sorry, will take a look during the week!

BTW don't you think we should do a release (containing the runtime arguments feature you added) first before adding this? The runtime arguments stuff is pretty huge already but is still in develop waiting.
Or maybe right after this PR, but I just hope I won't once again take too much time reviewing it before merging (swamped with SwiftGen + personal projects once again).

ilyapuchka:
Actually I have one more feature ready in a queue - auto-injections, that I also thought to suggest for next release.

AliSoftware:
Mmmh then maybe we should make a release right now with runtime parameters, then merge circular deps + auto-injections next, to make incremental releases instead of big leaps?

ilyapuchka:
I would let you to decide =) Maybe auto-injection feature will take some time to approve and merge in.

Dip/Dip/Definition.swift
```diff
@@ -50,20 +50,62 @@ func ==(lhs: DefinitionKey, rhs: DefinitionKey) -> Bool {
...
   /// Indicates that a new instance of the component will be created each time it's resolved.
   case Prototype
+  /// Indicates that instances will be reused during resolve but will be discurded when topmost `resolve` method returns.
+  case ObjectGraph
   /// Indicates that resolved component should be retained by container and always reused.
   case Singleton
 }
 
 ///Definition of type T describes how instances of this type should be created when they are resolved by container.
 public final class DefinitionOf<T>: Definition {
```
AliSoftware: I was thinking (didn't think it through, that's just a though), couldn't `DefinitionOf`be a `struct` to leverage value type?
Then `inScope` would return a new `DefinitionOf` instead of changing it in place (which seems less FP)?

Dip/Dip/Dip.swift
```diff
@@ -42,8 +42,7 @@ public class DependencyContainer {
     case Int(IntegerLiteralType)
   }
   
-  private var dependencies = [DefinitionKey : Definition]()
-  private var lock: OSSpinLock = OS_SPINLOCK_INIT
```
AliSoftware: Why removing the lock? (I added it to ensure thread-safety at that time)

Dip/Dip/Dip.swift
```diff
@@ -161,37 +167,74 @@ public class DependencyContainer {
...
+    func storeResolvedInstance<T>(instance: T, forKey key: DefinitionKey?, definition: DefinitionOf<T>) {
+      self.resolvedInstances[key] = instance
+      definition.resolvedInstance = instance
+    }
+    
+    func previouslyResolved<T>(key: DefinitionKey?, definition: DefinitionOf<T>) -> T? {
+      return (definition.resolvedInstance ?? self.resolvedInstances[key]) as? T
+    }
+    
+    var depth: Int = 0
```
AliSoftware: `private`?

Dip/Dip/Dip.swift
```diff
@@ -161,37 +167,74 @@ public class DependencyContainer {
...
+
+    func incrementDepth() {
```
AliSoftware: Wondering if it wouldn't be better to make a scoping-function, instead of a function pair. like:

```
incrementDepth(@noescape block: Void->Void) {
  depth++
  block()
  depth--
  ...
}
```

README.md
```diff
@@ -147,6 +147,20 @@ func resolve<T, Arg1, Arg2, Arg3, Arg4, Arg5, Arg6, Arg7>(tag tag: Tag? = nil, _
 
 ```
 
+### Circular dependencies
```
AliSoftware: Probably add a page for that in the playground as well?

Dip/Dip/Definition.swift
ilyapuchka: That will not work because if those methods on `DefinitionOf` will produce new copies then we will need to replace them in registry. We could solve that by having `register<T>(definition: DefinitionOf<T>)` and creating definition manually, but that will break initially nice syntax in separate steps without adding anything valuable.

Dip/Dip/Dip.swift
ilyapuchka: If we have a lock inside `resolve` method then it will deadlock if you try to call `resolve` inside factory closure or `resolveDependenciesBlock` of definition.

Dip/Dip/Definition.swift
AliSoftware: Ok I see that now, but tbh I'm not sure I like that "nice syntax" of being able to `register { … }.inScope(.ObjectGraph)`. Because having that `inScope` operator being able to change an already-registered dependency means that one can then change it at any time even later after registration, which feels wrong to me:
```
let def = container.register { [unowned container] ClientImp(server: container.resolve() as Server) as Client }
def.inScope(.ObjectGraph)
// … use it
def.inScope(.Singleton)
// … use it again
def.inScope(.Prototype)
// ... use it again
```

Dip/Dip/Dip.swift
AliSoftware: Yeah but removing it breaks thread-safety. Better replace it with a `NSRecursiveLock` instead maybe?

Dip/Dip/Dip.swift
ilyapuchka: Then it should be like this:

```swift
private func _resolve<T, F>(key: DefinitionKey?, definition: DefinitionOf<T>, builder: F->T) -> T {
  return resolvedInstances.resolve {
    ...
  }
}

...

func resolve<T>(@noescape block: ()->T) -> T {
  depth++
  let resolved = block()
  depth--
  if depth == 0 {
    resolvedInstances.removeAll()
  }
  return resolved
}
```

Looks more complex then just `defer`...

AliSoftware: No, you will just have to replace:
```
resolvedInstances.incrementDepth()
defer { resolvedInstances.decrementDepth() }
// rest of the code
```
with:
```
resolvedInstances.incrementDepth {
  // rest of the code
}
```

What I don't like about the function pair is that it's stateful. You can call one and forget to call the other and don't balance your calls. With a scoping function you can't make that mistake and you can't end up in an intermediate, unfinished state, you're guaranteed to be balanced, by construct.

Dip/Dip/Definition.swift
ilyapuchka: That's fair. What about `resolveDependencies` block?

Dip/Dip/Dip.swift
ilyapuchka: As for me using any kind of lock is bad practice and leads to more hard to catch bugs... 
Pretty silly example but still possible: someone call resolve not from main thread and then for any reason switch back to another thread in factory block and call `resolve` on that thread. It will still create a deadlock as I understand. So it looks like we exchange one set of "you do not do that this way" to another.

But if you still would like to use locks we can give it a try. Though I think there is no need to lock `register` method cause they are supposed to be called at one point in a program. The only place worth locking is probably access to resolved instances pool. What do you think?

Dip/Dip/Definition.swift
AliSoftware: Haven't grasped that one yet. It seems like it has the same issue (mutating postfix function), but not sure how to avoid it in this case or if that's as problematic.
I guess I'm less annoyed by this case somehow. Maybe we could add a security in that `resolveDependencies` method so that it `fatalError` if we try to call it while it already has a value and so has already been called before, to prevent multiple calls at least?

Dip/Dip/Dip.swift
ilyapuchka: Agree about safety. But here we need to return result of resolve from the method so the scoping function block should return what is produced by "result of the code".

Dip/Dip/Dip.swift
AliSoftware: Yeah, seems safe to only lock on `resolve` as registrations are usually all done in one place at the start indeed.

But generally the philosophy in multi-threaded environments and when implementing thread-safety is to protect the access to a resource. When you add a lock to project a resource, you wrap _all_ the accesses to that resource with the lock, otherwise by principle the lock won't prevent _all_ concurrent access, so feels strange to not respect that usual pattern and only protect _some_ accesses to `dependencies` and not _not_… I'm hesitating.

(Anyway if we decide to either remove the lock or only lock resolutions and not registrations, that would be a breaking change that would need to be documented in the README and CHANGELOG)

Dip/Dip/Definition.swift
ilyapuchka: But if that will be still method on a `struct` then it will mutate it or produce new instance instead of affecting one that is in registry. We could make it an optional argument on `register` method, but that will break using factory as trailing closure (don't know if you care about that)

Dip/Dip/Dip.swift
AliSoftware: Then what about implementing it this way?

```swift
func incrementDepth<T>(@noescape block: Void->T) -> T {
  depth++
  let r = block()
  depth--
  if  depth == 0 {
    resolvedInstances.removeAll()
  }
  return r
}
```

or even if you prefer to use `defer`:
```swift
func incrementDepth<T>(@noescape block: Void->T) -> T {
  depth++
  defer {
    depth--
    if  depth == 0 {
      resolvedInstances.removeAll()
    }
  }
  return block()
}
```

What I mean is just instead of expecting the call site to call `increment` + `defer { decrement }`, move all that inside the `increment` function anyhow.

Dip/Dip/Definition.swift
AliSoftware: Yeah I agree and I do care about being able to use the trailing closure syntax, feels nice there.

I think I'm ok with:
* Removing `inScope` method and instead make it a parameter on `register`
* Either make `DefinitionOf` a `struct` + make `resolveDependencies` `mutating`
* Or keep `DefinitionOf` a `final class`
* In either case, make it `fatalError` if `resolveDependenciesBlock` is already non-`nil` when you call it, meaning that it's the second time you call it
* In such context, I'm wondering if it's still worth it to `return self` there as you won't probably need chaining with that new configuration, so better not tempt the user to chain somehow (besides, I don't like chaining mutating functions, I think mutating functions should be their own instructions)

What do you think?

Dip/Dip/Dip.swift
ilyapuchka: My initial point was that it less error prone to ensure single thread access to shared resource (container in our case) by client then to make container itself guaranty thread safety in all possible cases. Just document that `resolve` should be used from single thread (probably the main?). Really I think it will be very rare case when user will call `resolve` from somewhere else then main thread.

AliSoftware: Ok fine by me then. Just document it somewhere in README & CHANGELOG ;-)

Dip/Dip/Dip.swift
ilyapuchka: Actually without `defer` in `incrementDepth` it looks more straight forward to me =) Also `resolve` (or something like that) instead of `incrementDepth` looks like more descriptive for me, cause `incrementDepth` implies that there should be also decrement depth. Anyhow agree that this is better.

AliSoftware: Yeah didn't like keeping the `incrementDepth` method name if moving to a scoping function, I agree that we need a more suitable name for that method then (maybe `resolve` indeed?).

AliSoftware:
I have to go, will let you make the pending changes and I'll take a deeper look directly in Xcode later

Dip/Dip/Definition.swift
ilyapuchka: We can try this:
- make `DefinitionOf<T, F>` where F will be type of factory and use it for factory property instead of `Any`
- add container argument to `resolveDependencies` method of `DefinitionOf` and use it to reinsert updated definition
- add `remove<T, F>(_: DefinitionOf<T, F>)` and `register<T, F>(_: DefinitionOf<T, F>)` to container (these methods can be handy anyway)

If that would not work (I suspect there will be ambiguity in method calls caused by generics) I would stick to using `final class`, adding fatal error and removing `return self` in `resolveDependencies`.

AliSoftware: Yeah, I think in that case directly keeping `final class` + fatalError + remove `return self` is probably the way to go. Adding that `<T,F>` will probably make the code harder to read — and indeed there is a also a risk of ambiguity for the compiler so better keep it simple and go with your second solution right away.

ilyapuchka: Actually it worked and I kind of satisfied with result. You can see it in late commit for this PR.
Using additional generic type added some complexity for sure, but that also lets get rid of force casting factory (which was `Any`) in `resolve` method.
Another two very positive side effects of this refactoring is ability to register/remove individual dependencies. Without generic type of factory that would be impossible cause there will be no way to compare definitions (because factory is of type 'Any', so no way to use `==`).
Drawback is that now you have to separate creating definition and call to resolve dependencies like this:

```
    var serverDefinition = container.register(.ObjectGraph) { Server() as Server }
    serverDefinition.resolveDependencies(container) { container, server in
      server.client = container.resolve() as Client
    }
```

If you are ok with that change I will proceed with updating readme and playground.

AliSoftware: Can't we do that?
```
container.register(.ObjectGraph) { Server() as Server }
  .resolveDependencies(container) { container, server in
    server.client = container.resolve() as Client
  }
```
Sure it might seem odd to chain a call after a trailing closure, but that works, and some might be ok with that style. I myself as a user of the lib I would prefer writing it like you did with two lines but others can choose to chain too, so I'm ok with that ;)

One thing newcomers might be wondering is why you call `resolveDependencies` on `serverDefinition` and not the `clientDefinition` of your ObjectGraph or both. Be sure to explain the rationale and logic of that in the playground.

ilyapuchka: If `resolveDependency` is mutating method then this will not work cause methods return immutable values (the same as using `let serverDefinition`  instead of `var serverDefinition`). But I guess we can make it not mutating and return new definition. Made another commit with that change.

ilyapuchka:
@AliSoftware Updated playground and documentation. I think it's ready for merge now.

ilyapuchka:
Now I've tried to reimplement auto-injection feature on top of that and it looks like structs semantics of `DefinitionOf` makes it much more harder. Maybe it would be better to rollback to `final class` variant.

AliSoftware:
I've finally decided to release the 3.0.0 first, as the rest of the PRs will be big changes anyway and I liked having intermediate release. Even if we'll hopefully merge this PR and the next one soon, feel better to go incremental.

I've renamed the "Sample app" folder to "SampleApp" (I like avoiding spaces in paths), so you might want to rebase your branches on top of the new develop. (I was gonna do it myself but as it will mess with your local git working copy I didn't want to surprise you)

---

Regarding the auto-injection feature being harder to implement because of `DefinitionOf` being a `struct` instead of a `final class`, I'm ok in keeping it a `final class` in the end as long as we have as few mutating functions as possible in the API (like avoid methods like the previous `inScope` one)

Will take a look at the playground now :wink: and keep you posted.

Dip/DipTests/ComponentScopeTests.swift
```diff
@@ -0,0 +1,125 @@
...
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+    super.tearDown()
+  }
+  
+  func testThatPrototypeIsDefaultScope() {
+    let def = container.register { ServiceImp1() as Service }
+    XCTAssertEqual(def.scope, ComponentScope.Prototype)
+  }
+  
+  func testThatCallingInScopeChangesScope() {
```
AliSoftware: This test is mis-named now that `inScope` is not a thing anymore :wink:

Dip/Dip/Dip.swift
```diff
@@ -77,24 +97,41 @@ public class DependencyContainer {
...
+   Pass `nil` to associate with any tag.
    - parameter instance: The instance to register, with return type of protocol you want to register it for
    
    - note: You must cast the instance to the protocol you want to register it with (e.g `MyClass() as MyAPI`)
+   
+   **Deprecated**: Use `register(.Singleton){}` method instead to define singleton scope.
    */
-  public func register<T>(tag tag: Tag? = nil, @autoclosure(escaping) instance factory: ()->T) -> DefinitionOf<T> {
-    return register(tag: tag, factory: { factory() }, scope: .Singleton)
+  @available(*, deprecated, message="Use `register(.Singleton){}` method instead to define singleton scope.")
```
AliSoftware: :cry: I liked the use of `@autoclosure` here, that's not every day we have a reason to use it :stuck_out_tongue_winking_eye: 

Too bad I didn't see that depreciation before, it would have made sense to include it in 3.0.0 so that we could remove the method altogether in the next 4.0.0.

If the next release changes the major version (if the API of your next PRs change from the API of 3.0.0 and isn't backwards compatible anymore) maybe we will do a 3.1.0 that just adds this depreciation warning before publishing 4.0.0 so we can completely remove that method in 4.0.0 then?

Dip/Dip/Definition.swift
```diff
@@ -50,29 +50,70 @@ func ==(lhs: DefinitionKey, rhs: DefinitionKey) -> Bool {
 public enum ComponentScope {
   /// Indicates that a new instance of the component will be created each time it's resolved.
   case Prototype
+  /// Indicates that instances will be reused during resolve but will be discurded when topmost `resolve` method returns.
+  case ObjectGraph
   /// Indicates that resolved component should be retained by container and always reused.
   case Singleton
 }
 
 ///Definition of type T describes how instances of this type should be created when they are resolved by container.
```
AliSoftware: Now that this has two generic parameters, maybe better describe them in the comments, like "`T` is the type of the instance to resolve to, `F` is the type of the factory to generate that instance, namely the runtime parameters that needs to be given when resolving that instance" or sthg similar.

AliSoftware:
I finally did the rebase myself and merged this PR after review.

I admit I didn't have time to review all your code implementation to double-check it, I figured I trust you on this and better not wait too much before merging.

Be careful as this as been rebased now: you'll probably need to rebase your new `auto-injection` on top of `develop`.
Feel free to insert some commits to revert `DefinitionOf` back to a `final class` before rebasing, if that `struct` still bothers you for the implementation of `auto-injection`. I'm ok with it as long as we don't make methods that mutate the object if not necessary.

AliSoftware:
Side note: just realized that now that `register(tag:instance:)` is deprecated, the Sample App has depreciation warnings all over :wink:

ilyapuchka:
Thanks for very constructive review! =)

AliSoftware:
Well maybe I merged a bit early as I forgot to address the comment about the documentation of `<T,F>` maybe you could improve that on `develop`?
(It's a small thing but with the library growing more and more and its implementation being more complex than in the beginning it's important to keep track :wink:)

Dip/Dip/Dip.swift
ilyapuchka: Yes, let's do that. The reason why I made this method deprecated was to keep consistency in how you register components in different scopes. But I'm not against keeping it if you want, so feel free to revert that.

AliSoftware: No I think it's a good thing to deprecate it, makes the API more consistent as you say, just sad to see it go :D

PR #19, opened 2015-11-26
Named first runtime argument in resolve

### LINKED ISSUE #18: Make first runtime argument parameter named
With introduction of runtime arguments (#8) the API became a bit confusing cause of mixing tags and arguments values. It would be better to use something like this: 

```swift
container.register(tag: "tag") { (arg1: Arg1) in ... as Protocol }
container.resolve(tag: "tag", withArguments: arg1) as Protocol
```

To release this feature faster and without bumping major version (what we are going to do to release circular dependencies) we can leave old version and later (or can we do it right away?) deprecate it.

ilyapuchka:
Resolves #18

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 #### New Features
 
-* Added support for circular dependencies by adding `ObjectGraph` scope to reuse resolved instances.  
+* Added support for circular dependencies:
+    * Added `ObjectGraph` scope to reuse resolved instances 
+    * Added `resolveDependencies` method on `DefinitionOf` class to resolve dependencies of resolved instance.  
   [#11](https://github.com/AliSoftware/Dip/pull/11), [@ilyapuchka](https://github.com/ilyapuchka)
 * Added methods to register/remove individual definitions.  
   [#11](https://github.com/AliSoftware/Dip/pull/11), [@ilyapuchka](https://github.com/ilyapuchka)
@@ -14,6 +16,7 @@
 * Removed container thread-safety to enable recursion calls to `resolve`.  
   **Access to container from multiple threads should be handled by clients** from now on.
 * Deprecated `register(tag:instance:)` method in favor of `register(.Singleton, …)`.
+* Added name for the first runtime argument in `resolve(tag:withArguments:)` methods to make more clear separation between tag and factory runtime arguments.
 
 ## 3.0.0
 
--- a/Dip/Dip/RuntimeArguments.swift
+++ b/Dip/Dip/RuntimeArguments.swift
@@ -56,6 +56,10 @@ extension DependencyContainer {
    
    - seealso: `resolve(tag:builder:)`
    */
+  public func resolve<T, Arg1>(tag tag: Tag? = nil, withArguments arg1: Arg1) -> T {
+    return resolve(tag: tag) { (factory: (Arg1) -> T) in factory(arg1) }
+  }
+  
   public func resolve<T, Arg1>(tag tag: Tag? = nil, _ arg1: Arg1) -> T {
     return resolve(tag: tag) { (factory: (Arg1) -> T) in factory(arg1) }
   }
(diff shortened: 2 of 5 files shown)


AliSoftware:
This means a 3.1.0 before the 4.0.0, right?

ilyapuchka:
I don't mind bumping major version and release it together with other things that are in develop now. To release 3.1 first will be too much updates in short time. But do you think it will be ok just to remove old methods without deprecating them first? Currently in develop `register(instance)` is only deprecated. Then we can remove them all together in 4.1?

P.S. probably I don't have access rights to create milestones and labels...

AliSoftware:
I think removing without deprecating might be maybe harsh but not _that_ problematic either, as long as we provide a clear note in the CHANGELOG.

We can't deprecate in 4.0.x and remove in 4.1.x because that would break [semantic versioning](http://semver.org): when you remove a method and thus break the backward compatibility you have to bump the major version. So the solutions to chose from are:

* deprecate it in 3.1 and remove in 4.0
* deprecate in 4.0 and remove in 5.0
* remove in 4.0 without deprecating it before

I think the first one would be the best, because typically when you do semver you use `~> 3.0` — so a `pod update` will update to 3.1 but not 4 until you change your Podfile manually. But if we dont plan to release a 3.1, then the last one won't be as problematic as if we were Apple or a big lib IMHO. After all we only have 50 stars so far, not 1k :wink:

ilyapuchka:
Ok, I see. But then if we deprecate it in 3.1 then we need to provide an alternative in the same version, don't we? For named arguments we have it, but for registering singleton we need to extract it from circular dependencies implementation. If it's ok I can do that.

AliSoftware:
Indeed, I didn't realize that the alternative for `register(instance:)` was only provided by the Circular Dependencies PR. To follow semver and be nice citizen with the API changes we should indeed extract the alternate `register(.Singleton)` from there so it can be included in 3.1 before we release Circular Dependencies in 4.0.

ilyapuchka:
Ok, let's do that. I think it will be easier to start `release/3.1.0` brunch from current master and do it there (we can't start it from develop cause circular dependencies are already merged in). And we need to close this then two.

AliSoftware:
Yeah we messed up a bit with te branches this time, it wasn't proper gitflow as we didn't have the roadmap yet, woops.

We can still create the branch from develop, but from an earlier commit in develop before we merged circular dependencies PR (and cherry-pick the missing, more recent commits from develop into release/3.1.0 if needs be). I think it would feel better than starting a release from master.

ilyapuchka:
Ok, will do that tomorrow.

AliSoftware:
Well I just took a look at the git graph and realize that… I already did exactly that 4 days ago :smile: 

Branch `origin/release/3.1.0` [was created 4 days ago](https://github.com/AliSoftware/Dip/branches) and points the the commit just after the release 3.0.0 and just before the Circular Dependencies branches started :wink:

ilyapuchka:
So can we close that?

ilyapuchka:
Maybe it will also worth adding `inScope` name to scope parameter in `register`? It can be also confusing if you use it together with tag: 
`register(tag: "tag", .Singleton) { ... }` vs `register(tag: "tag", inScope: .Singleton) { ... }` or `register(inScope: .Singleton) { ... }`

AliSoftware:
Mmh given that the Scope is a fixed enum, I'm not sure the confusion can be done like there was the confusion between tag and runtime argument before?

ilyapuchka:
Yep, let's leave it like it is.
So I've made a pull request for 3.1.0 release - https://github.com/AliSoftware/Dip/pull/21

AliSoftware:
Cool! Will close this issue once I took time to merge #21 :)

ilyapuchka:
We now don't need this as it is now in develop after merge of release/3.1.0 - 97782364167a1d8504d730922f31176ab3683229

PR #22, opened 2015-11-30
Throwing errors

### LINKED ISSUE #15: Throwing errors instead of fatalError
Probably we should consider throwing errors in `resolve` instead of using `fatalError`. For now we use it only in one place, but there could be more places where we will need to signal developer that something is wrong. `try` enforces developer to pay attention to possible errors. Also with throwing errors we can test those invalid cases what we can't do with `fatalError`.

ilyapuchka:
Resolves #15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,14 @@
   [#11](https://github.com/AliSoftware/Dip/pull/11), [@ilyapuchka](https://github.com/ilyapuchka)
 * Added methods to register/remove individual definitions.  
   [#11](https://github.com/AliSoftware/Dip/pull/11), [@ilyapuchka](https://github.com/ilyapuchka)
+* All `resolve` methods now can throw error if type can not be resolved.  
+  [#15](https://github.com/AliSoftware/Dip/issues/15), [@ilyapuchka](https://github.com/ilyapuchka)
 
 #### Breaking Changes
 
 * Removed container thread-safety to enable recursion calls to `resolve`.  
   **Access to container from multiple threads should be handled by clients** from now on.
+* All `resolve` methods now can throw.
 
 ## 3.1.0
 
--- a/Dip/Dip/Dip.swift
+++ b/Dip/Dip/Dip.swift
@@ -96,7 +96,7 @@ public class DependencyContainer {
   container.register(tag: "service") { ServiceImp() as Service }
   container.register(.ObjectGraph) { ServiceImp() as Service }
   container.register { [unowned container] 
-    ClientImp(service: container.resolve() as Service) as Client 
+    ClientImp(service: try! container.resolve() as Service) as Client
   }
   ```
   */
@@ -152,13 +152,16 @@ public class DependencyContainer {
   If no definition was registered with this `tag` for this `protocol`,
   it will try to resolve the definition associated with `nil` (no tag).
   
+  Will throw `DipError.DefinitionNotFound` if no registered definition found
+  that would match type, runtime arguments and tag.
+  
   - parameter tag: The arbitrary tag to look for when resolving this protocol.
   
   **Example**:
   ```swift
-  let service = container.resolve() as Service
-  let service = container.resolve(tag: "service") as Service
-  let service: Service = container.resolve()
+  let service = try! container.resolve() as Service
+  let service = try! container.resolve(tag: "service") as Service
+  let service: Service = try! container.resolve()
   ```
   
   */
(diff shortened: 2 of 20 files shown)


CHANGELOG.md
```diff
@@ -8,13 +8,15 @@
     * Added `ObjectGraph` scope to reuse resolved instances 
     * Added `resolveDependencies` method on `DefinitionOf` class to resolve dependencies of resolved instance.  
   [#11](https://github.com/AliSoftware/Dip/pull/11), [@ilyapuchka](https://github.com/ilyapuchka)
-* Added methods to register/remove individual definitions.  
+* Added methods to register/remove individual definitions
   [#11](https://github.com/AliSoftware/Dip/pull/11), [@ilyapuchka](https://github.com/ilyapuchka)
+* All `resolve` methods now can throw error if type can not be resolved [#15](https://github.com/AliSoftware/Dip/issues/15), [@ilyapuchka](https://github.com/ilyapuchka)
```
AliSoftware: Little nitpicking :wink: The convention we started for CHANGELOG is to end the line describing the change with `.  ` (dot followed by two spaces) then mention the issue/PR number + author on the next line.

```
* Describe the change.␣␣
  [#n](…), [@author](…)
```

Dip/Dip/Dip.swift
```diff
@@ -178,21 +181,20 @@ public class DependencyContainer {
...
    */
-  public func resolve<T, F>(tag tag: Tag? = nil, builder: F->T) -> T {
+  public func resolve<T, F>(tag tag: Tag? = nil, builder: F->T) throws -> T {
     let key = DefinitionKey(protocolType: T.self, factoryType: F.self, associatedTag: tag)
     let nilTagKey = tag.map { _ in DefinitionKey(protocolType: T.self, factoryType: F.self, associatedTag: nil) }
 
     guard let definition = (self.definitions[key] ?? self.definitions[nilTagKey]) as? DefinitionOf<T, F> else {
-      fatalError("No definition registered with " + (tag == nil ? "\(key)" : "\(key) or \(nilTagKey)") + ". "
-        + "Check the tag, type you try to resolve, number, order and types of runtime arguments passed to `resolve()`.")
+      throw DipError.DefinitionNotFound(key)
```
AliSoftware: This is indeed way better, I like that change :+1:

Dip/DipTests/ComponentScopeTests.swift
```diff
@@ -90,14 +90,14 @@ class ComponentScopeTests: XCTestCase {
 
   func testThatItReusesInstanceInObjectGraphScopeDuringResolve() {
     //given
-    container.register(.ObjectGraph) { [unowned container] in Client(server: container.resolve()) as Client }
+    container.register(.ObjectGraph) { [unowned container] in Client(server: try! container.resolve()) as Client }
     
     container.register(.ObjectGraph) { Server() as Server }.resolveDependencies { container, server in
```
AliSoftware: Not related to this PR I know, but now that I read that again I'm wondering: shouldn't we use `[unowned container]` (or at least `[weak container]`) here? It seems to me that `container` strongly reference `resolveDependenciesBlock` which itself strongly references `container`, hence a strong reference cycle, right?

ilyapuchka: I don't think so, cause here we don't capture container from outer context, we pass it as an argument at runtime. So the reference should be released as soon as block returns. I suppose...

Speaking of `unowned` I tried to use factory block without `[unowned container]` it (like in line 93) and it still looks like there is no retain cycle... I test it by having a strong reference to container and a weak reference. When I release a strong reference weak also is being released. That would not happen in retain cycle... but if I think of it there should be one if there is no `[unowned container]` or `[weak container]`, cause container holds definitions and they hold their factories... Maybe it's just too late at night to think about that :sleepy:  :night_with_stars:

AliSoftware: > I don't think so, cause here we don't capture container from outer context, we pass it as an argument at runtime. So the reference should be released as soon as block returns.

Don't we? I mean, `container` captures `resolveDependenciesBlock` right? So unless that block don't use the `container` argument in the closure, this `container` is captured automatically by the closure and not "released as soon as the block returns" (as the `resolveDependenciesBlock` will still be captured and kept as a property in the `container`). So this captured `container` will only be released when the block itself is released (set back to `nil`).

(If the block were not stored as a strong property by the container but used right away, we would be able to mark it `@noescape`, which is not the case here)

ilyapuchka: I've created separate issue #23 to discuss that. Will probably add some unit test for that.

AliSoftware: :+1:

PR #26, opened 2015-12-07
Added targets for OS X, tvOS and watchOS2

### LINKED ISSUE #4: Support watchOS2, tvOS, OSX?
I don't think anything should prevent `Dip` from working on those other platforms, but I'll have to check first by creating some example before adding that support in the `podspec`.

ilyapuchka:
Resolves #4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ install:
 script:
 - set -o pipefail && xcodebuild test -workspace Dip.xcworkspace -scheme Dip-iOS -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest' ONLY_ACTIVE_ARCH=NO | xcpretty -c
 - set -o pipefail && xcodebuild test -workspace Dip.xcworkspace -scheme Dip-OSX -sdk macosx -destination 'platform=OS X,arch=x86_64' ONLY_ACTIVE_ARCH=NO | xcpretty -c
-- set -o pipefail && xcodebuild test -workspace Dip.xcworkspace -scheme Dip-tvOS -sdk appletvsimulator ONLY_ACTIVE_ARCH=NO | xcpretty -c
-- set -o pipefail && xcodebuild -workspace Dip.xcworkspace -scheme Dip-watchOS -sdk watchsimulator ONLY_ACTIVE_ARCH=NO | xcpretty - c
+- set -o pipefail && xcodebuild test -workspace Dip.xcworkspace -scheme Dip-tvOS -sdk appletvsimulator -destination 'platform=tvOS Simulator,name=Apple TV 1080p,OS=latest' ONLY_ACTIVE_ARCH=NO | xcpretty -c
+- set -o pipefail && xcodebuild -workspace Dip.xcworkspace -scheme Dip-watchOS -sdk watchsimulator -destination 'platform=watchOS Simulator,name=Apple Watch - 38mm,OS=latest' ONLY_ACTIVE_ARCH=NO | xcpretty - c
 - set -o pipefail && xcodebuild test -workspace Dip.xcworkspace -scheme DipSampleApp -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest' ONLY_ACTIVE_ARCH=NO | xcpretty -c
 - pod lib lint --quick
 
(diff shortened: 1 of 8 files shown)


AliSoftware:
If the `Info.plist` files are really all exactly the same, you might want to avoid duplicating them, and instead use the same for all targets — simply specifying the same file in each target's `INFO_PLIST_FILE`  (or sthg similar) Build Setting

AliSoftware:
Only reviewing from GitHub, but apart from the remark above, looks good from there, so if you confirm that all the targets compiles and all the unit tests pass and all those targets, I'm :+1: 

Don't forget the entry in the CHANGELOG though.

AliSoftware:
When targeting iPhone Simulator (= for running tests on travis), you shouldn't CodeSign the bundle, that will avoid code signing issues on Travis-CI.
Code Signing will only be needed when building the framework as a final product to consume it (to release it to others).

You might be able to disable Code Signing from the command line by passing the Build Settings via environment variables, like `CODE_SIGNING_IDENTITY=- xcodebuild … test` and similar.

ilyapuchka:
it just does not work :angry: 
btw I also updated a podspec

ilyapuchka:
That option should be fine (or it should be CODE_SIGNING_IDENTITY=""), it just should work without it as worked before and still works for sample app...

ilyapuchka:
Now it works. Sorry for all that late night notifications

AliSoftware:
So what did you change to make it work? (You seem to have collapsed all the commits into one so I can't see your commit history :-/)

It looks like you disabled code signing directly in the Build Settings of the xcodeproj project? If so, does this still allow you to `carthage build --no-skip-current` and build the framework even if not signed? (we might want to add this commands to `.travis.yml` btw to make the CI ensure it doesn't break either)

ilyapuchka:
Yes, I just disabled code signing for OS X target. Can you please check if carthage builds them for you? On my machine it was working fine.

AliSoftware:
Uh oh.

```
$ carthage build --no-skip-current Dip
*** xcodebuild output can be found in /var/folders/j8/2wg5p4dx6yg54_fj5yn1rp280000gn/T/carthage-xcodebuild.bC6Dnj.log
*** Building scheme "Dip-watchOS" in Dip.xcodeproj
** BUILD FAILED **


The following build commands failed:
	Check dependencies
(1 failure)
Code Sign error: No code signing identities found: No valid signing identities (i.e. certificate and private key pair) matching the team ID “87BXKFRW6K” were found.
A shell task failed with exit code 65:
** BUILD FAILED **
```

AliSoftware:
Also is it intentional that the Dip-watchOS framework target doesn't have any associated Unit Test target?

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
 * All `resolve` methods now can throw error if type can not be resolved.  
   [#15](https://github.com/AliSoftware/Dip/issues/15), [@ilyapuchka](https://github.com/ilyapuchka)
 * `DependencyContainer` is marked as `final`.
+* Added supprot for OSX, tvOS and watchOS2.  
+[#26](https://github.com/AliSoftware/Dip/pull/26), [@ilyapuchka](https://github.com/ilyapuchka)
```
AliSoftware: * typo: `supprot`--> `support`
* nitpicking: add two spaces at beginning of the `[#26]` line, so it's aligned with the line above

.travis.yml
```diff
@@ -9,7 +9,8 @@ install:
 - gem install xcpretty --no-rdoc --no-ri --no-document --quiet
 
 script:
-- set -o pipefail && xcodebuild test -workspace Dip.xcworkspace -scheme Dip -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest' ONLY_ACTIVE_ARCH=NO | xcpretty -c
+- set -o pipefail && xcodebuild test -workspace Dip.xcworkspace -scheme Dip-iOS -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest' ONLY_ACTIVE_ARCH=NO | xcpretty -c
+- set -o pipefail && xcodebuild test -workspace Dip.xcworkspace -scheme Dip-OSX -sdk macosx -destination 'platform=OS X,arch=x86_64' ONLY_ACTIVE_ARCH=NO | xcpretty -c
 - set -o pipefail && xcodebuild test -workspace Dip.xcworkspace -scheme DipSampleApp -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest' ONLY_ACTIVE_ARCH=NO | xcpretty -c
```
AliSoftware: We should also add scripts for `Dip-watchOS` and `Dip-tvOS` schemes in the `.travis.yml`

ilyapuchka:
Sorry, didn't notice that also committed my personal team id. Regarding disabling code signing it's disabled only for OS X test target, I hope travis will not mind that.
Regarding tests target for watchOS2 there is no option to create test target for watchOS in Xcode.

AliSoftware:
Ok I fixed the Code Signing issue with my own commit.
Before closing this PR we should probably add an entry on `.travis.yml` for tvOS scheme. And maybe an entry to just _build_ the watchOS scheme if there is no way to add tests for them (seems strange though)

AliSoftware:
Wait, what did you do your last commit for, I had it working :laughing:

ilyapuchka:
Didn't noticed it, sorry. Removed my commit.

.travis.yml
ilyapuchka: Ok, I will do that. Probably you will see some notifications about failed builds =)

ilyapuchka:
I've also added `carthage build --no-skip-current`step

AliSoftware:
:ok_hand:

PR #31, opened 2015-12-19
Add thread safety

mwoollard:
Add locking around resolution and dictionary modification to ensure thread safety

My understanding is Dip isn't currently thread safe when resolving, with risk of two instances of singleton being returned if same type requested at same time from different threads. How about this change to address this? I'd ideally want any container I used to be thread safe :-)

Let me know what you think !

Kind regards
Mark

### DIFF
--- a/Dip/Dip/Dip.swift
+++ b/Dip/Dip/Dip.swift
@@ -41,28 +41,33 @@ public final class DependencyContainer {
   }
   
   var definitions = [DefinitionKey : Definition]()
+  var lock: NSRecursiveLock?
   
   /**
    Designated initializer for a DependencyContainer
    
    - parameter configBlock: A configuration block in which you typically put all you `register` calls.
-   
+   - parameter isThreadSafe: If true instance is thread safe, if false not thread safe but
+       slight performance gain. Default is to be thread safe.
    - note: The `configBlock` is simply called at the end of the `init` to let you configure everything. 
            It is only present for convenience to have a cleaner syntax when declaring and initializing
            your `DependencyContainer` instances.
    
    - returns: A new DependencyContainer.
    */
-  public init(@noescape configBlock: (DependencyContainer->()) = { _ in }) {
+  public init(isThreadSafe: Bool = true, @noescape configBlock: (DependencyContainer->()) = { _ in }) {
+    if isThreadSafe {
+      lock = NSRecursiveLock()
+    }
     configBlock(self)
   }
   
   // MARK: - Thread safety
   func threadSafe(closure: () -> Void) {
     
-    objc_sync_enter(self)
+    lock?.lock()
     defer {
-      objc_sync_exit(self)
+      lock?.unlock()
     }
     closure()
     
@@ -70,9 +75,9 @@ public final class DependencyContainer {
   
   func threadSafe<T>(closure: () -> T) -> T {
     
-    objc_sync_enter(self)
+    lock?.lock()
     defer {
-      objc_sync_exit(self)
+      lock?.unlock()
     }
     return closure()
     
--- a/Dip/Dip/Dip.swift
+++ b/Dip/Dip/Dip.swift
@@ -41,45 +41,40 @@ public final class DependencyContainer {
   }
   
   var definitions = [DefinitionKey : Definition]()
-  var lock: NSRecursiveLock?
+  let lock = NSRecursiveLock()
   
   /**
    Designated initializer for a DependencyContainer
    
    - parameter configBlock: A configuration block in which you typically put all you `register` calls.
-   - parameter isThreadSafe: If true instance is thread safe, if false not thread safe but
-       slight performance gain. Default is to be thread safe.
-   - note: The `configBlock` is simply called at the end of the `init` to let you configure everything. 
+   - note: The `configBlock` is simply called at the end of the `init` to let you configure everything.
            It is only present for convenience to have a cleaner syntax when declaring and initializing
            your `DependencyContainer` instances.
    
    - returns: A new DependencyContainer.
    */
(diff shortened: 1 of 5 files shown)


AliSoftware:
We removed thread safety in the latest release because it lead to deadlocks when doing circular dependencies resolution (see CHANELOG)

I want to restore this thread safety as it is important but we have to be careful it doesn't generate issues with circular dependencies as demonstrated in the playground and add enough unit tests to ensure this in long term.

mwoollard:
OK - thanks for the feedback

AliSoftware:
In your case I'm not sure `objc_sync_enter` and `objc_sync_exit` are reentrant and maybe we'll need to switch to NSRecursiveLock or something but then we'll have to study performance impact.

ilyapuchka:
@mwoollard what's your use case where you need container to be thread safe?
As @AliSoftware said we need to make sure it does not dead lock on circular dependencies. Did you run tests?

mwoollard:
@AliSoftware objc_sync_enter is recursive according to the comments in its definition header, stating it uses a recursive pthread_mutex internally. @ilyapuchka the current unit tests all passed. I can look adding some tests for thread safety if you'd like. Can run some performance comparisons with and without locking too. No specific use case right now as am just starting looking at IoC /w swift but multi-threaded access is likely as I move forward and just something I noticed.

mwoollard:
So tests suggest NSRecursiveLock is around 4 times more efficient than objc_sync_enter/exit (testing on 2 GHz Intel Core i7 MacBook Air). Using this as the lock approx doubles resolve time overhead (same system). Have committed some tests for threading and timing and added param to DependencyContainer init to allow thread safety to be turned off if desired.

Dip/Dip/Dip.swift
```diff
@@ -201,30 +233,30 @@ public final class DependencyContainer {
...
   private func _resolve<T, F>(tag: Tag? = nil, key: DefinitionKey?, definition: DefinitionOf<T, F>, builder: F->T) -> T {
     
-    return resolvedInstances.resolve {
-      
-      if let previouslyResolved: T = resolvedInstances.previouslyResolved(key, definition: definition) {
-        return previouslyResolved
-      }
-      else {
-        let resolvedInstance = builder(definition.factory)
+    return threadSafe {
```
ilyapuchka: Maybe it will be better to lock only whole `resolve<T, F>(tag tag: Tag? = nil, builder: F->T) throws -> T ` method that calls this one? The same will apply to other methods that will call `_resolve` in future (if there will be any). As for me less locking operations is better.

Dip/DipTests/ThreadSafetyTests.swift
```diff
@@ -0,0 +1,150 @@
...
+        lock.unlock()
+      }
+    }
+    
+    queue.waitUntilAllOperationsAreFinished()
+    
+    XCTAssertEqual(resultSet.count, 100)
+  }
+  
+  func testCircularReferenceThreadSafety() {
```
ilyapuchka: I would also test it in a different way - buy switching thread inside `resolveDependencies` block of one of the definitions. Even if it's not likely to be done directly there could be cases when client code uses some method in this closure that will switch thread and at some point call another `resolve`.

ilyapuchka:
I think this thread safety parameter is an overkill. Dependency resolution is not intended to be used to resolve so many instances that it can drastically decrease performance compared to not using locks and that users would like to turn it off. For me it just makes api and implementation (optional lock) weird. 
Also I would not include benchmark tests for in Dip tests. They are good for reference, but I believe are not needed every time.
I also have some other minor comments regarding code and testing circular dependencies, commented it in code.
Apart from that looks fine, thanks for pull request, @mwoollard !

Dip/Dip/Dip.swift
mwoollard: Yes, thinking about it a bit more, makes sense - less locking overhead

mwoollard:
Yes, thread safety option not needed, was added to compare performance

Dip/DipTests/ThreadSafetyTests.swift
mwoollard: Have added some thread switching within resolveDependencies block in each threading test

ilyapuchka: Sorry, I meant something different, like this:

```swift
container.register(.ObjectGraph) { Server() as Server }.resolveDependencies { container, server in
  var client: Client
  dispatch_sync(dispatch_get_global_queue()) {
    client = try! container.resolve() as Client
  }
  server.client = client
}
```
The idea is to check that `client.service` is resolved correctly even if `client` and `service` are resolved on different threads.
And I meant to do it in a separate test, other tests don't need that.

mwoollard: Ah, right. I did try resolving on separate thread but without the sync wait in the dependency closure, that doesn't work as each resolve triggers another as resolved items array cleared before other thread gets to it :-) Latest should be good !

ilyapuchka:
@AliSoftware your move, should we merge it?

Dip/Dip/Dip.swift
```diff
@@ -57,13 +57,35 @@ public final class DependencyContainer {
     configBlock(self)
   }
   
+  // MARK: - Thread safety
+  private func threadSafe(closure: () -> Void) {
+
```
AliSoftware: For code style coherence, please remove leading and trailing empty lines inside both those `threadSafe` implementations.

AliSoftware:
Apart from the nitpicking at code styling, LGTM :+1: 

Don't forget to add an entry in `CHANGELOG.md` for that change and to credit @mwoollard though :wink:
Nice job!

AliSoftware:
Before merging this there is two remaining things:

* An entry in the CHANGELOG
* Edit the `README.md` to remove / alter the paragraph talking about Dip not being thread-safe

PR #32, opened 2016-01-10
Throwing factories and resolve dependencies block

ilyapuchka:
With that PR we no more need to call `container.resolve()` with `try!` inside a factory or `resolveDependencies` block. Also convenient to use throwing constructors or any other throwing methods inside those blocks. Error will be catched by container, printed in console and then will be re-thrown.

Note: There is no way (at least for now) I can think of to be able to distinguish between throwing and not-throwing factories (thus the latest definition will override previous), except of duplicating all `register`/`resolve` methods, what I don't like at all cause it will double the number of methods in container. Taking in account that tag is an optional argument and we have support for 6 runtime arguments already (and upcoming multi-injection feature will ad 6 more methods) it will make API too much bloated.

### DIFF
--- a/Dip/Dip/Dip.swift
+++ b/Dip/Dip/Dip.swift
@@ -97,7 +97,7 @@ public final class DependencyContainer {
   container.register { ClientImp(service: try! container.resolve() as Service) as Client }
   ```
   */
-  public func register<T>(tag tag: Tag? = nil, _ scope: ComponentScope = .Prototype, factory: ()->T) -> DefinitionOf<T, ()->T> {
+  public func register<T>(tag tag: Tag? = nil, _ scope: ComponentScope = .Prototype, factory: () throws -> T) -> DefinitionOf<T, () throws ->T> {
     return registerFactory(tag: tag, scope: scope, factory: factory)
   }
   
@@ -163,7 +163,7 @@ public final class DependencyContainer {
   
   */
   public func resolve<T>(tag tag: Tag? = nil) throws -> T {
-    return try resolve(tag: tag) { (factory: ()->T) in factory() }
+    return try resolve(tag: tag) { (factory: () throws -> T) in try factory() }
   }
   
   /**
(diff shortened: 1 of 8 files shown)


Dip/Dip/Dip.swift
```diff
@@ -186,28 +186,30 @@ public final class DependencyContainer {
    Though before you do that you should probably review your design and try to reduce the number of dependencies.
    
    */
-  public func resolve<T, F>(tag tag: Tag? = nil, builder: F->T) throws -> T {
+  public func resolve<T, F>(tag tag: Tag? = nil, builder: F throws -> T) throws -> T {
```
AliSoftware: Not sure if the outer `throws` should be `throws` or `rethrows`?

http://robnapier.net/re-throws

Dip/Dip/Dip.swift
```diff
@@ -186,28 +186,30 @@ public final class DependencyContainer {
...
     }
 
     let usingKey: DefinitionKey? = definition.scope == .ObjectGraph ? key : nil
-    return _resolve(tag, key: usingKey, definition: definition, builder: builder)
+    return try _resolve(tag, key: usingKey, definition: definition, builder: builder)
   }
   
   /// Actually resolve dependency
-  private func _resolve<T, F>(tag: Tag? = nil, key: DefinitionKey?, definition: DefinitionOf<T, F>, builder: F->T) -> T {
+  private func _resolve<T, F>(tag: Tag? = nil, key: DefinitionKey?, definition: DefinitionOf<T, F>, builder: F throws -> T) throws -> T {
```
AliSoftware: (Same open question here)

Dip/Dip/Dip.swift
ilyapuchka: Here `rethrows` will not compile cause this method throws error itself.

Dip/Dip/Dip.swift
ilyapuchka: But here yes, `rethrows` will work.

AliSoftware:
Just make sure you double-check every time you use `(f: … throws) -> throws` if it's valid or if a `rethrows` might instead be used. But apart from that LGTM.

AliSoftware:
Note: once you merge that, make sure to rebase #13 before merging it, as I'm wondering if this change might have impacts on the playground / example code you added there too, like changing calls using `try` there?

ilyapuchka:
Looks like `_resolve` is the only place where `rethrows` is applicable, except `ResolvedInstances.resolve` where it is already used.

AliSoftware:
:ok_hand:

ilyapuchka:
Good catch, these changes actually broke #13 =)

ilyapuchka:
Fixed the issue, it safe to merge this now :tada:

PR #35, opened 2016-01-15
Auto-injection improvements

ilyapuchka:
With this PR I try to make property auto-injection rules more strict thus safer. 

Generally property injection is less safe than constructor injection and should be used mostly in situations where dependency is optional because there is a good default implementation. Swift actually makes property injection almost unneeded with default arguments values. But still there could be cases when property injection is more correct than constructor injection (i.e. when instance is always created not by user code, but by system - that can be view controller created by storyboard).

I think that properties that should be auto-injected by Dip should be managed and created only by container. Otherwise user can replace underlying value with manually created instance that can be not fully constructed (if it also uses property injection). So it will be safer if they would be readonly for client code. We can not prevent defining properties as `var`s but we can make `value` property of wrapper readonly, making it impossible to create wrapper by any means except container.

This also makes auto-injected properties required readonly properties and make sure that dependency will be always set, or resolve would fail completely. Originally error was caught and Dip moved to the next property.

It also reduces chances of coupling with Dip by preventing user from using auto-injection wrappers anywhere but property definition.

Also in this PR I add blocks to wrappers that behave like `didSet` observers for standard Swift properties. If user already have some property injected manually with `didSet` observer he can use this block in the same way when/if he decide to use auto-injection for that property.

If users will find these rules too strict and not practical we can change that later without breaking backward compatibility.

### DIFF
--- a/Dip/Dip.xcodeproj/project.pbxproj
+++ b/Dip/Dip.xcodeproj/project.pbxproj
@@ -38,6 +38,10 @@
 		0919F4EC1C16419500DC3B10 /* DefinitionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0919F4CF1C16417000DC3B10 /* DefinitionTests.swift */; };
 		0919F4ED1C16419500DC3B10 /* RuntimeArgumentsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0919F4D21C16417000DC3B10 /* RuntimeArgumentsTests.swift */; };
 		0919F4EE1C16419500DC3B10 /* ComponentScopeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0919F4CE1C16417000DC3B10 /* ComponentScopeTests.swift */; };
+		0982AF0C1C5183A000B62463 /* Utils.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0982AF0B1C5183A000B62463 /* Utils.swift */; };
+		0982AF0D1C5183A000B62463 /* Utils.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0982AF0B1C5183A000B62463 /* Utils.swift */; };
+		0982AF0E1C5183A000B62463 /* Utils.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0982AF0B1C5183A000B62463 /* Utils.swift */; };
+		0982AF0F1C5183A000B62463 /* Utils.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0982AF0B1C5183A000B62463 /* Utils.swift */; };
 		09873F561C1E0237000C02F6 /* AutoInjection.swift in Sources */ = {isa = PBXBuildFile; fileRef = 09873F551C1E0237000C02F6 /* AutoInjection.swift */; };
 		09873F771C1E024E000C02F6 /* AutoInjectionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 09873F751C1E0249000C02F6 /* AutoInjectionTests.swift */; };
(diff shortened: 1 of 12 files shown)


ilyapuchka:
@AliSoftware what are your fillings about that? Is making auto-injected properties required is too strict? I'm in doubt now regarding this aspect...

By it's nature property injection should be used in situations where dependency is _optional_ cause there is a local default implementation in place (or no dependency at all is ok). In iOS we have a special case of storyboards that create objects for us, so instead of using constructor injection we will have to use property injection even if dependency is required. We still can use standard property injection and `resolveDependencies`. But auto-injection becomes less applicable in general. The thing is that there is currently only one way - either to treat auto-injected properties as always optional or always required.

One option I see here is to add argument to `Injected` constructor like this: `init(required: Bool = false, didInject: T -> () = { _ in })` and check it at runtime. I will note ahead that I have already implemented tagged auto-injection so that if we do so users will be able to define property like this: `var server = Injected<Server>(tag: "default", required: true)`. The other option is to have separate set of `OptionalInjected<T>`/`OptionalWeakInjected<T>` wrappers which will be treated the same way except error handling.

Much swifty solution will be probably like using `Injected<Server>` for required and `Injected<Server?>` for optional property, but that seems hard to resolve internally by Dip, without making user to register `Server` and `Server?` separately (even with that there will be probably different problems).

Will be glad to hear your input on that =)

ilyapuchka:
Soooo, now there are much more great improvements in this PR. It turns out that my original implementation on auto-injection, that relied on custom tagged definitions, was over-complicated, and much easier solution is possible. You will see now that implementation is much cleaner and streamline.
Also I've added support for tagged auto-injection, with new implementation it's just a piece of cake.

@AliSoftware if you will have any problems reviewing this, just let me know. I know that PR probably became to big for review to be easy

AliSoftware:
Could you explain to me the difference between `autoInjectProperties` & `resolveDependencies`? Not sure the difference would be clear for a library consumer?

ilyapuchka:
The difference is that `resolveDependenciesOf` searches the definition but instead of calling it's factory it uses passed in instance. Then it will call `resolveDependenciesBlock` of that definition and then `autoInjectProperties`.
I've added `resolveDependenciesOf` to support storyboards integration. Most likely it's the only valid scenario to use this method.
Also I've made it generic to be able to specify a protocol that ViewController can implement (if user uses protocols for view controllers too).

AliSoftware:
Ok.

It's probably worth mentioning that in the function's comment (especially
that the most likely scenario is for Storyboard injection for in this one)
and in the playground too maybe.

My fear when I see Dip growing is that adding too much features and
especially too much functions in the public API will make the library
confusing and scary for any newcomer, so better explain each function the
best we could and with practical examples, because for a lot of people DI
is scary and hard to get 😉

Le dim. 31 janv. 2016 à 22:29, Ilya Puchka <notifications@github.com> a
écrit :

ilyapuchka:
Actually now when you asked I see that this method probably does not belong to Dip, but should be better moved to DipUI to make it clear that it's intended to be used with storyboards.

AliSoftware:
So if I understand correctly, [this new `autoInjectProperties()` function](https://github.com/AliSoftware/Dip/pull/35/files#diff-0ab526653b6b3b6fe83ccd63c1dadef7R32) the new name / replacement of [the former function `resolveDependencies()` which has been removed in this PR here](https://github.com/AliSoftware/Dip/pull/35/files#diff-0ab526653b6b3b6fe83ccd63c1dadef7L144), right?

ilyapuchka:
Yes, exactly.
Also notice that I make it not public. This way it will be just the final step in standard resolve process. And to call it user will need to register the type of the instance he wants to inject to in the container. In original implementation the idea was to let user to auto-inject properties in any instance, whether it is registered or not in the container. The reason for that change is that I want to enforce using container to manage and create all the instances, so it will be aware of them. And separately we handle the special case of storyboard instantiation. Also this will decrease confusion and enforce the same pattern of using `resolve`, instead of adding another method that only does some small part of it.
If there will be any need we can make this method public in future release.

AliSoftware:
Perfect then!

PR #37, opened 2016-01-27
Refined sample app and readme

ilyapuchka:
I took a chance to refine sample code and readme to make it cleaner, and particularly making it free of not recommended practice of referencing container throughout the code base.


SampleApp/DipSampleApp/Providers/FakePersonsProviders.swift
```diff
@@ -59,3 +86,28 @@ class PlistPersonProvider : PersonProviderAPI {
         )
     }
 }
+
+class FakePersonsProvider: PersonProviderAPI {
+    
+    let dummyProvider: PersonProviderAPI
+    let plistProvider: PersonProviderAPI
+    
+    init(dummyProvider: PersonProviderAPI, plistProvider: PersonProviderAPI) {
```
AliSoftware: I think it could be interesting to add a comment on that `init` to explain that this is also DI: a lot of people don't realize that they are doing "constructor injection" here and that Dependency Injection could sometimes be simple as that.

The first time I learned about DI it scared me and when I realized that simply passing my dependencies as parameter to the constructor instead of creating them inside the class implementation was also a form of DI, that helped be understand and realize that DI was not that complex after all.

So I think it's worth mentioning as a comment, here and on all the other `init(…)` around the Example project where we do constructor injection too.

AliSoftware:
Apart from the additional comments I mentioned above, I reviewed the rest and it seems :ok_hand: to me.
So once you've added the comments on the `init` using constructor injection, feel free to merge this PR yourself :+1:

ilyapuchka:
I've added comments you have asked for. Also I refactored code a bit to demonstrate use of tagged definitions, runtime arguments and property injection.

PR #39, opened 2016-01-28
Multiple-injection

### LINKED ISSUE #14: Multi-injection
Sometimes it can be useful to register several implementations for the same protocol and resolve them as an array. Imagine that you have different authorisation services, for instance via Facebook, Twitter and Google, that all implement a common protocol, and you want to present the list of all possible services so that user can choose one of them but your code does not care what concrete service he chose, you just work with it via that common protocol. 

With Dip we can do that only with tags, associating different implementations of the same protocol with different tags and resolve them using the same tags. Though it works it introduces coupling between independent parts of the system - point of registration and points of resolution. 

```swift
protocol OAuthService {
    func doLogin()
}

class FacebookOAuth: OAuthService {
    func doLogin() {
        print("login with facebook")
    }
}

class TwitterOAuth: OAuthService {
    func doLogin() {
        print("login with twitter")
    }
}

class GoogleOAuth: OAuthService {
    func doLogin() {
        print("login with google")
    }
}

let container = DependencyContainer { container in
    container.register(tag: "facebook") { FacebookOAuth() as OAuthService }
    container.register(tag: "twitter") { TwitterOAuth() as OAuthService }
    container.register(tag: "google") { GoogleOAuth() as OAuthService }
}

let services: [OAuthService] = [
    container.resolve(tag: "facebook"),
    container.resolve(tag: "twitter"),
    container.resolve(tag: "google")
]

services[Int(arc4random_uniform(UInt32(3)))].doLogin()
```

Other IoC containers (like [Ninject](http://www.ninject.org) for .Net) has this feature. The obvious consequence of allowing multiple bindings to the same protocol is that how you resolve them not as an array but as a single instance, what binding (definition in case of Dip) you should choose. Ninject offers different styles of attributes or runtime context checks that are very specific for .Net. If there is no constraints that resolution result is ambiguous and exception is thrown.

Do we need such feature in Dip and if yes than how we should solve the ambiguity problem?

ilyapuchka:
Resolves #14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,11 @@
 * Added auto-injection feature.  
   [#13](https://github.com/AliSoftware/Dip/pull/13), [@ilyapuchka](https://github.com/ilyapuchka)
 * Factories and `resolveDependencies` blocks of `DefinitionOf` are now allowed to `throw`.  
-  [#32](https://github.com/AliSoftware/Dip/pull/13), [@ilyapuchka](https://github.com/ilyapuchka)
+  [#32](https://github.com/AliSoftware/Dip/pull/32), [@ilyapuchka](https://github.com/ilyapuchka)
 * Thread safety reimplemented with support for recursive methods calls.  
   [#31](https://github.com/AliSoftware/Dip/pull/31), [@mwoollard](https://github.com/mwoollard)
+* Multi-injection to resolve an array of all the instances registere for the protocol.  
+  [#39](https://github.com/AliSoftware/Dip/pull/39), [@ilyapuchka](https://github.com/ilyapuchka)
 
 
 ## 4.0.0
--- a/Dip/Dip/Dip.swift
+++ b/Dip/Dip/Dip.swift
@@ -35,7 +35,7 @@ public final class DependencyContainer {
    with the same protocol, to differentiate them. Tags can be either String
    or Int, to your convenience.
    */
-  public enum Tag: Equatable {
+  public enum Tag: Equatable, Comparable {
     case String(StringLiteralType)
     case Int(IntegerLiteralType)
   }
@@ -120,7 +120,7 @@ public final class DependencyContainer {
   container.register { ClientImp(service: try! container.resolve() as Service) as Client }
   ```
   */
-  public func register<T>(tag tag: Tag? = nil, _ scope: ComponentScope = .Prototype, factory: () throws -> T) -> DefinitionOf<T, () throws ->T > {
+  public func register<T>(tag tag: Tag? = nil, _ scope: ComponentScope = .Prototype, factory: () throws -> T) -> DefinitionOf<T, () throws -> T> {
     return registerFactory(tag: tag, scope: scope, factory: factory)
   }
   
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -7,9 +7,11 @@
 * Added auto-injection feature.  
   [#13](https://github.com/AliSoftware/Dip/pull/13), [@ilyapuchka](https://github.com/ilyapuchka)
 * Factories and `resolveDependencies` blocks of `DefinitionOf` are now allowed to `throw`.  
-  [#32](https://github.com/AliSoftware/Dip/pull/13), [@ilyapuchka](https://github.com/ilyapuchka)
+  [#32](https://github.com/AliSoftware/Dip/pull/32), [@ilyapuchka](https://github.com/ilyapuchka)
 * Thread safety reimplemented with support for recursive methods calls.  
   [#31](https://github.com/AliSoftware/Dip/pull/31), [@mwoollard](https://github.com/mwoollard)
+* Multi-injection to resolve an array of all the instances registere for the protocol.
```
AliSoftware: `registered`

Dip/Dip/Dip.swift
```diff
@@ -340,14 +385,43 @@ extension DependencyContainer.Tag: StringLiteralConvertible {
   }
 }
 
+/*
+.String("a") == .Stirng("a")
```
AliSoftware: `String`

Dip/Dip/Dip.swift
```diff
@@ -340,14 +385,43 @@ extension DependencyContainer.Tag: StringLiteralConvertible {
...
+/**
+ Tags are compared by comparing strings created from their assiciated values.
+ 
+ .String("a") < .String("b")
+ .Int(0) < .Int(1)
+ .String("0") < .Int(1)
+ .Int(0) < .String("1")
+ 
+ */
+public func <(lhs: DependencyContainer.Tag, rhs: DependencyContainer.Tag) -> Bool {
```
AliSoftware: This seems like a strange addition. I've not finished reviewing the PR yet so maybe I'm gonna see why this is needed later in the code, but I'm not sure why you need to compare tags and make them `Comparable` here. Checking for Equality makes sense, sorting tags, not so sure…?

ilyapuchka: You are right, looks like it leaked from some WIP for other feature. I will remove that.

AliSoftware:
Already said that in #14 but I'm really not sure this is worth adding.

I like having a library that doesn't become a huge do-it-all lib. Keeping the API simple is important to me, and adding convenience methods to do something that the user can really easily do on its own if really needed does not seem to go in that direction, but rather adds methods in the API for which the user might wonder why it's here (in the sense "that feature is so simple to do it on my own, why is there a dedicated method to do this, if there is a dedicated method does it means that this is in fact not as obvious as I think it is?" etc)


In your examples, it seems to me that using a simple `map` allows to achieve the exact same effect that this feature does:

```swift
enum Services: String {
  case GMail, Yahoo, Outlook
  let allValues: [Services] = [GMail, Yahoo, Outlook]
}
container.register(tag: Services.GMail) { GmailService() as ThirdPartyEmailService }
container.register(tag: Services.Yahoo) { YahooService() as ThirdPartyEmailService }
container.register(tag: Services.Outlook) { OutlookService() as ThirdPartyEmailService }

let thirdPartyServices = try! Services.allValues.map { container.resolve(tag: $0) as ThirdPartyEmailService }
```

ilyapuchka:
Sure, it can be done by user with ease. I'm ok with not including that feature.

PR #41, opened 2016-02-06
Swift package manager support

### LINKED ISSUE #30: Create a Package.swift
To early support the Swift Dependency Manager

ilyapuchka:
Resolves #30


AliSoftware:
I think it would be the perfect occasion to _move_ the `Dip/Dip` folder to `Sources` (Instead of just symlinking)?

ilyapuchka:
Yes, if you prefer that. I tried both, moving files and symlink - both works.

AliSoftware:
Cool! Just don't forget to update the podspec too once you moved the source files :+1:

PR #42, opened 2016-02-06
Implemented recursive lock using pthread mutex

ilyapuchka:
Would be cool if Dip will not depend on Foundation. The only dependency now is NSRecursiveLock. So I grabbed implementation from [here](https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/Concurrency/Lock.swift)

Still needs to be tested. Will try it later on virtual machine, but need to set it up first.

### DIFF
--- a/Dip/DipLinuxTests/Sources/main.swift
+++ b/Dip/DipLinuxTests/Sources/main.swift
@@ -1,0 +1,156 @@
+import Glibc
+import XCTest
+import Dip
+
+
+private protocol Service: class {
+  var client: Client? { get set }
+}
+
+private protocol Client: class {
+  var service: Service { get }
+}
+
+private class ClientImp: Client, Equatable {
+  var service: Service
+  init(service: Service) {
+    self.service = service
+  }
+}
+
+private func ==<T: ClientImp>(lhs: T, rhs: T) -> Bool {
+  return lhs === rhs
+}
+
+private class ServiceImp: Service, Hashable {
+  weak var client: Client?
+  init() {}
+  
+  var hashValue: Int {
+    return unsafeAddressOf(self).hashValue
+  }
+}
+
+private func ==<T: ServiceImp>(lhs: T, rhs: T) -> Bool {
+  return lhs === rhs
+}
+
+
+typealias TMain = @convention(c) (UnsafeMutablePointer<Void>) -> UnsafeMutablePointer<Void>
+
+func dispatch_async(block: TMain) {
+  var pid: pthread_t = 0
+  pthread_create(&pid, nil, block, nil)
+}
+
+func dispatch_sync(block: TMain) -> UnsafeMutablePointer<Void> {
+  var pid: pthread_t = 0
+  var result: UnsafeMutablePointer<Void> = nil
+  pthread_create(&pid, nil, block, nil)
+  pthread_join(pid, &result)
+  return result
+}
+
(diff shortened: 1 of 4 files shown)


AliSoftware:
Love the idea :+1:

ilyapuchka:
Managed to implement thread safety tests for Linux :tired_face: I think it will be not very hard to port all other tests. Will do that later in a separate PR.

```
["Test Case \'ThreadSafetyTests.testSingletonThreadSafety\' started."]
["Test Case \'ThreadSafetyTests.testSingletonThreadSafety\' passed (1.004 seconds)."]
["Test Case \'ThreadSafetyTests.testFactoryThreadSafety\' started."]
["Test Case \'ThreadSafetyTests.testFactoryThreadSafety\' passed (1.007 seconds)."]
["Test Case \'ThreadSafetyTests.testCircularReferenceThreadSafety\' started."]
["Test Case \'ThreadSafetyTests.testCircularReferenceThreadSafety\' passed (2.008 seconds)."]
["Executed 3 tests, with 0 failures (0 unexpected) in 4.019 (4.027) seconds"]
["Total executed 3 tests, with 0 failures (0 unexpected) in 4.019 (4.029) seconds"]
```

Dip/DipLinuxTests/Sources/main.swift
```diff
@@ -0,0 +1,156 @@
...
+      dispatch_async { _ in
+        let client = try! container.resolve() as Client
+        pthread_spin_lock(&lock)
+        resolvedClients.append(client as! ClientImp)
+        pthread_spin_unlock(&lock)
+        return nil
+      }
+    }
+    
+    sleep(2)
```
AliSoftware: why?

Dip/DipLinuxTests/Sources/main.swift
```diff
@@ -0,0 +1,156 @@
...
+  pthread_create(&pid, nil, block, nil)
+}
+
+func dispatch_sync(block: TMain) -> UnsafeMutablePointer<Void> {
+  var pid: pthread_t = 0
+  var result: UnsafeMutablePointer<Void> = nil
+  pthread_create(&pid, nil, block, nil)
+  pthread_join(pid, &result)
+  return result
+}
```
AliSoftware: Shouldn't both of those be in `Utils.swift`?

ilyapuchka: That's only a test code. I did't have a chance to checkout what is there in swift-core-libdispatch, if it is usable I will try to replace it.

Dip/DipLinuxTests/Sources/main.swift
ilyapuchka: Just to let al the threads finish. Of course joining each thread would be better (then dispatch_xxx should return thread's pid). Again it's only because of limitations of swift-core libs, in this case xctest does not provide async utils yet.

AliSoftware:
I'll trust you on this one and let you merge this when you're ready, because I won't have time to install a Linux VM on my Mac to test the code in there :wink:

ilyapuchka:
Ok =) I really hope it will work. At least I was able to fix the issue with `lock` vs `tryLock` (first one was causing deadlock).

I'm still struggling to understand how to organize Linux tests better, for instance how to avoid putting all of them in one file, but I will address that in a separate PR.

PR #43, opened 2016-02-07
Singleton scope fixed

ilyapuchka:
The problem is that instances managed using Singleton scope should not be shared between containers. But currently there is a way to register the same definition in different containers, so if we store resolved instance in definition it will be shared between those two containers. In this PR I moved storing such instances in resolved instances pool of the container that is not drained when resolve method returns. So the instances will be properly scoped to containers.


AliSoftware:
Shouldn't we add Unit Tests to cover this case and avoid further regression?

AliSoftware:
LGTM. You can merge right after fixing that typo :+1:

ilyapuchka:
Thanks for review! :wink:

PR #46, opened 2016-02-13
Building tests on linux with spm

### LINKED ISSUE #44: Port tests to Linux
#42 ported only thread safety tests. Need to port all other tests too. Will probably need to refactor a bit existing tests to make them compatible. With that Dip will be "Linux ready".

ilyapuchka:
In this PR I've added support to run tests on Linux, to be sure Dip is properly working on that platform too.
Resolves #44


ilyapuchka:
Didn't encounter any issues so far, so I'm going to close this PR and prepare a release :tada: 
Ideally we should run it on travis, but I'm not sure if it supports running on different platforms

PR #50, opened 2016-02-23
DependencyTagConvertible for better typed Tags

gavrix:
This offers to remove `Tag` from `register` and `resolve` APIs  in favour of `DependencyTagConvertible` protocol, which may represent Tag itself as well as any other type conforming to this protocol. 
Specifically, this allows me to use custom enums (conforming to `DependencyTagConvertible`) as tags instead of `String` or `Int` literals, making it more type-safe.

I modified SampleApp to use such an enum for tags. [see here](https://github.com/AliSoftware/Dip/compare/develop...gavrix:develop#diff-f678440884b69096a2fcf0b90b04c486R61)

At the same time, it is still possible to use `String` and `Int` literals as tags, so this PR adds more flexibility retaining existing functionality

### DIFF
--- a/SampleApp/DipSampleApp/DependencyContainers.swift
+++ b/SampleApp/DipSampleApp/DependencyContainers.swift
@@ -17,6 +17,17 @@ private let FAKE_STARSHIPS = false
 /* ---- */
 
 
+enum DependencyTags: Int {
+    case Hardcoded
+    case Dummy
+}
+
+extension DependencyTags: DependencyTagConvertible {
+    func toTag() -> DependencyTag {
+        return DependencyTag.Int(self.rawValue)
+    }
+}
+
 // MARK: Dependency Container for Providers
 func configureContainer(dip: DependencyContainer) {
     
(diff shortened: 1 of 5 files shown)


AliSoftware:
I think we discussed this possibility at some point with @ilyapuchka 

I was in favor of using a more generic type, but in the end using protocols lead to introducing PATs (Protocol with Associated Types) which led to a bunch of issues for some use cases, like you can't create an array of PATs in Swift by design…

Have to recheck that now that the codebase evolved, maybe the issue with PATs is not that string anymore, but I know we thought about it and had a reason not to do it at that time…

AliSoftware:
Ah found it: see discussion in #3

ilyapuchka:
Thanks for PR, @gavrix! 

Actually I personally [gave up](https://github.com/AliSoftware/Dip/wiki/Roadmap#version-410-soon-1) on that feature. It is kind of convenient, but does not bring any significant value except passing some type (probably enum case) without calling `tag` or `toTag()` or whatever on the instance of it. But at the same time it complicates the API adding one more abstraction (tag convertible protocol).

I didn't run the code, but it looks for me like it's a breaking change - before we had `DependencyContainer.Tag`, now - `DependencyTag`. Is it? If we add this feature I would personally prefer not to bump the version just for the sake of it. That's my main concern about it. Even if we move `Tag` enum from `DependencyContainer` scope (but I would prefer to keep it there - I'm not a huge fan of that, just to not bump version) I would prefer to keep the name as `Tag` (in case of ambiguity users can use it like `Dip.Tag`). Having `DependencyTag` and `DefinitionKey` types can cause questions. And regarding `toTag` method - having `tag` property looks more consistent with `CustomStringConvertible` protocol.

gavrix:
Looks like I need to fix some tests to these changes.

The main idea why I think this is a better API is that from API user perspective I have no idea how tags are organized inside Dip, but I have to provide it either a string or int literal, or a custom conversion to DependencyContainer.Tag. 

Literals are extremely unsafe: you can make typo, you can break something during refactoring without knowing it etc. Everyone on swift tries to replace raw literals usage with types, mostly enums.

Regarding associated types — yeah, known problem, the solution is using type erasure technique. But my implementation doesn't use associated types.

While is it still possible to have all possible tags listed in enum in my client code and have function converting that enum to `DependencyContainer.Tag` instance, uneducated users will continue using raw literals (btw, following examples you gave) producing low quality swift code. 

Basically what I propose is to integrate that function I mentioned (converting my custom tags enum to `DependencyContainer.Tag`) into Dip. It is probably worth removing possibility to provide raw literal as a tag instance altogether (I intentionally preserved that possibility).

ilyapuchka:
@gavrix can you please look into failing tests and fix them?
Also please add type alias like this:

```swift
class DependencyContainer {
  typealias Tag = DefinitionTag
}
```
This way the change will not break code that explicitly used `DependencyContainer.Tag`. If type aliases support `@available` attribute please also add it with a note that it was deprecated in favour of new type.

Regarding new type for tag I still don't feel `DependencyTag` is a good name. Maybe it would be better to use just `Tag` in global scope, or move it to `DefinitionKey` scope, so that the type would be `DefinitionKey.Tag`. Just `DefinitionTag` may also work, but it can cause confusion with `DefinitionKey`.

sergey-at-digiflare:
Sure, let me take a look at it.

gavrix:
Sure, let me take a look at it.

ilyapuchka:
@gavrix Thanks!

PR #61, opened 2016-03-22
Fix issues from Xcode 7.3 / Swift 2.2 upgrade

mwoollard:
Needed to make these changes to build with Swift 2.2. This will however break earlier versions of Swift so might not be what you want if you feel maintaining backward compatibility is needed. My initial thoughts on this was having to #if / #else / # endif and duplicate most of the code in this file to deal with change from __FILE__ / __LINE__ to #file / #line was inelegant !


ilyapuchka:
Looks like your develop is not up to date, 'cause I see some old commits in this PR that should not be there. Can you please look at that?
I will take a look on your fix closely in the evening. That's only tests so maybe it's not a big deal. Also I'm not sure if it will continue to work on Linux... Maybe will take some time to investigate cause I killed my virtual machine yesterday to free up some space to update Xcode...

mwoollard:
Will do - I did update my fork as per github instructions prior to fixing... I see that CI build now broken as Travis still on 2.1 which might be another issue for you depending on how quickly they update.

mwoollard:
Re-created fork to sort out merge and re-created PR on that fork

PR #62, opened 2016-03-22
Fix build warnings / issues from Swift 2.2

mwoollard:
Re-created the pull request to sort out merge issues. Note this PR breaks backward compatibility with Swift 2.1 and earlier and not validated on Linux.

### DIFF
--- a/Dip/DipTests/Sources/Utils.swift
+++ b/Dip/DipTests/Sources/Utils.swift
@@ -24,26 +24,20 @@
 
 import XCTest
 
-#if os(Linux)
-typealias FileString = StaticString
-#else
-typealias FileString = String
-#endif
-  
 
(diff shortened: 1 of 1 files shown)


ilyapuchka:
Probably we will need to keep a separate brunch for 2.1 for some time to make it work on Travis until they complete transition. Then we can drop support for 2.1, don't see any reason to keep it after that.

ilyapuchka:
As I can see these changes in __FILE__/__LINE__ are just deprecations, so let's keep them. I really don't want to break builds (even for some time) and don't want to keep a separate brunch for another language version.
Also I've got error that `try expression()` is ambiguous. Had to change it  to `try expression() as T`. Didn't you get the same error?

ilyapuchka:
Never mind, Travis already updated :tada:

PR #65, opened 2016-03-25
EagerSingleton scope and bootstrap method

### LINKED ISSUE #63: Allow for creating instances in the container right after it is initialized
This is mostly about the Singleton scope. In many cases, the "create and preserve after first resolve was made" approach can cause various issues, especially if some of these singletons would use events/signals for communication. To provide more flexibility, it would be great to provide a way for singletons to be created and preserved right away and not waiting for first resolve reference. This could be configured per singleton or as a general Dip property.

ilyapuchka:
Resolves #63 

New scope `EagerSingleton` lets users to register components that should be instantiated by container early in the app life cycle. To define this moment new method `bootstrap` was added to `DependencyContainer`, as was suggested by @sebastian-zarzycki-es. Method call is optional. But we can print warning in `resolve` method if it is called without calling bootstrap and warn that this method can become required in future versions. 
After bootstrap is called no modifications in registered definitions can be done (no removes, no adds). 

Obvious limitation of that approach is that it can only instantiate components using factories without runtime arguments. Should not be a serious issue though. And auto-wiring is still possible. If eager singleton has dependencies or somehow causes other singletons to be instantiated they also will be instantiated on bootstrap, even if they were not registered as eager singletons.

### DIFF
--- a/Dip/DipTests/Sources/ComponentScopeTests.swift
+++ b/Dip/DipTests/Sources/ComponentScopeTests.swift
@@ -96,78 +96,104 @@ class ComponentScopeTests: XCTestCase {
   }
   
   func testThatItReusesInstanceForSingletonScope() {
-    //given
-    container.register(.Singleton) { ServiceImp1() as Service }
+    func test(scope: ComponentScope) {
+      //given
+      container.register(scope) { ServiceImp1() as Service }
+      
+      //when
+      let service1 = try! container.resolve() as Service
+      let service2 = try! container.resolve() as Service
+      
+      //then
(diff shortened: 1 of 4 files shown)


mwoollard:
Wonder if BootstrapSingleton or InstantiatedSingleton may be better than EagerSingleton ?

ilyapuchka:
I chose "eager" because of the term "eager evaluation" - exactly what happens in this case, in opposite to default lazy evaluation.

mwoollard:
Ok - not heard that term before - just used to lazy as the alternative to normal initialisation:-)

PR #80, opened 2016-04-24
[WIP] Type forwarding and resolving optionals

ilyapuchka:
### Type forwarding
Sometimes we have situations when some component implements several protocols (provides several services). If we want to reuse the same instance for different services in the same object graph we had to do it manually, like this:

```swift
container.register { try ServiceImp(client: container.resolve()) as Service }
  .resolveDependencies { container, service in
    service.delegate = service.client as! ServiceDelegate
}
container.register { ClientImp() as Client }
```

or hack it around with this:

```swift
container.register { try ServiceImp(client: container.resolve()) as Service }
  .resolveDependencies { container, service in
    service.delegate = try container.resolve() as ServiceDelegate
}
container.register(.ObjectGraph) { ClientImp() as Client }
container.register { (try container.resolve() as Client) as! ServiceDelegate }
````

With type-forwarding we can register the same definition for several types and reuse produced instance in the same object graph:

```swift
container.register { try ServiceImp(client: container.resolve()) as Service }
  .resolveDependencies { container, service in
    service.delegate = container.resolve() as ServiceDelegate
}
container.register(.ObjectGraph) { ClientImp() as Client }
  .implements(ServiceDelegate.self)
```

This way we make our registrations more flexible, because later we can easily change definition for `ServiceDelegate` from `ClientImp` to something else without changing definition of service.

### Resolving optionals
Type forwarding also lets us easily resolve optional types. With type inference it is very easy to make a following mistake:

```swift
class Service {
  init(client: Client?)
}

container.register { ClientImp() as Client }

container.register { try ServiceImp(client: container.resolve()) as Service }
//or with auto-wiring
container.register { ServiceImp(client: $0) as Service }

//this will fail because container will try to resolve inferred `Client?` type which is not the same as `Client`
try! container.resolve() as Service
```

One way to solve the problem is to always specify type to resolve:

```swift
container.register { try ServiceImp(client: container.resolve() as Client) as Service }
container.register { ServiceImp(client: $0 as Client) as Service }
```

But that is easy to forget. With type forwarding Dip becomes smart enough to make it work. Under the hood  it implicitly registers optional types as forwarded types so no additional actions are required from the user.

```swift
container.register { try ServiceImp(client: container.resolve()) as Service }
//or with auto-wiring
container.register { ServiceImp(client: $0) as Service }

//will successfully resolve client
try! container.resolve() as Service
```
With that providing explicit types of dependencies in configurations becomes almost unneeded.

But it does not solve the problem of runtime arguments:

```swift
container.register { client in ServiceImp(url: client) as Service }

let client: Client = ...

//this will still fail because container registered `(Client?) throws -> Service` factory but will try to lookup `(Client) throws -> Service`
try! container.resolve(withArguments: client) as Service

let optClient: Client? = ...
//this will work
try! container.resolve(withArguments: optClient) as Service
```

#### Drawbacks:
- there is no way to ensure type-safety of `implements` method. If produced by definition instance does not implement forwarded type container will throw error
- if there are several definitions for different types registered with the same (or nil) tags and they both forward to the same type, container will throw error when resolving forwarded type because there is no way to select one of the definitions (without relying on internal implementation details of swift dictionary keys sorting).
- forwarded types will not override existing definitions for the same types and vice versa, what is different from standard registration behavior

### DIFF
--- a/Dip/DipTests/Sources/AutoWiringTests.swift
+++ b/Dip/DipTests/Sources/AutoWiringTests.swift
@@ -97,6 +97,30 @@ class AutoWiringTests: XCTestCase {
     XCTAssertTrue(anyClient is AutoWiredClientImp)
   }
   
+  func testThatItUsesTagToResolveDependenciesWithAutoWiring() {
+    //given
+    container.register(.ObjectGraph) { ServiceImp1() as Service }
+    container.register(tag: "tag", .ObjectGraph) { ServiceImp2() as Service }
+    container.register(.ObjectGraph) { ServiceImp2() }
+    
+    container.register(.ObjectGraph) { AutoWiredClientImp(service1: $0, service2: $1) as AutoWiredClient }
+    
+    //when
+    let client = try! container.resolve(tag: "tag") as AutoWiredClient
+    
+    //then
+    let service1 = client.service1
+    XCTAssertTrue(service1 is ServiceImp2)
+    let service2 = client.service2
+    XCTAssertTrue(service2 is ServiceImp2)
+    
+    //when
+    let anyClient = try! container.resolve(AutoWiredClient.self)
+    
+    //then
+    XCTAssertTrue(anyClient is AutoWiredClientImp)
+  }
+  
   func testThatItUsesAutoWireFactoryWithMostNumberOfArguments() {
     //given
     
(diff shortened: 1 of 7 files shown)


ilyapuchka:
It feels like drawbacks of current design of this feature are more crucial than the issue it solves. Will put this on hold for now. But it still can be used under the hood to resolve optionals - will create a separate pr for that.

ilyapuchka:
Implemented resolving optionals in #84 backed by type forwarding, but removed it from public interface to be able to revise it later.

PR #82, opened 2016-04-25
[WIP] Pass tag to auto-injected propertiies

ilyapuchka:
When auto-injecting properties we can pass tag used to resolve containing instance to provide a context and to be able to use tagged definition to resolve injected property. The same way tag is passed through to resolve dependencies with auto-wiring. If injected property is defined with explicit tag it will be used instead.


ilyapuchka:
This will be obsolete when #83 would be merged

PR #92, opened 2016-06-10
Enable the 'allow app extension api only' option

mwoollard:
Dip doesn't appear to use any API calls that are unavailable to app extensions. Enabling this flag removes linker warning in extension targets that link against dip.


ilyapuchka:
👍

PR #101, opened 2016-07-21
Updating to Swift 3.0

patrick-lind:
I pulled down the develop branch and converted it to Swift 3 (Or attempted to). It looks mostly successful with only a few of the unit tests failing still that I was hoping you could take a look at. I'm having some weirdness in Xcode that is making some of this debugging a challenge on the latest Xcode beta build. Let me know what you think!


ilyapuchka:
Thanks @patrick-lind I'l merge it and will take a look.

PR #105, opened 2016-07-28
Grand renaming

ilyapuchka:
With Swift3 API guidelines in mind here is what I propose to rename in public API:

1. `ObjectGraph` -> `Shared`, `Prototype` -> `Unique`. These names will make it more clear how these scopes affect instances
2. `resolveDependencies` -> `resolvingProperties`
3. `DefinitionKey` properties: `protocolType` -> `type`, `associatedTag` -> `tag`, `argumentsType` -> `typeOfArguments`. Additional words in original names do not add any value to understanding meaning of these properties.
4.  `resolve` method parameter `withArguments` -> `arguments`
5. switch places for `scope` and `tag` parameters in `register` method. This does not break API in Swift2 but with Swift 3 will require to follow the order from declaration (SE-0060). `register(.Shared, tag: "some")` looks nicer than `register(tag: "some", scope: .Shared)`

There are also some internal methods renaming, but obviously its not that critical.

### DIFF
--- a/Dip/DipTests/TypeForwardingTests.swift
+++ b/Dip/DipTests/TypeForwardingTests.swift
@@ -79,7 +79,7 @@ class TypeForwardingTests: XCTestCase {
   func testThatItReusesInstanceResolvedByTypeForwarding() {
     //given
     let def = container.register(.Shared) { ServiceImp1() as Service }
-      .resolveDependencies { container, resolved in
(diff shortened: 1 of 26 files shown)


ilyapuchka:
- [ ] `ComponentScope` enum cases should be lowercased to comply with swift 3 guidelines

PR #106, opened 2016-08-03
Updating to new Xcode 8 beta 4

patrick-lind:
Updating to reconcile with new Swift 3 changes made available in Xcode
8 beta 4

### DIFF
--- a/Sources/Utils.swift
+++ b/Sources/Utils.swift
@@ -143,4 +143,5 @@ extension Collection where Index: Comparable, Self.Indices.Index == Index {
   
 #else
   import Foundation
+  typealias RecursiveLock = NSRecursiveLock
 #endif
--- a/Dip/Dip.xcodeproj/project.pbxproj
+++ b/Dip/Dip.xcodeproj/project.pbxproj
@@ -436,6 +436,7 @@
 					};
 					0903B3AE1C1618AF002241C1 = {
 						CreatedOnToolsVersion = 7.1.1;
+						LastSwiftMigration = 0800;
 					};
 					0903B4031C162862002241C1 = {
 						CreatedOnToolsVersion = 7.1.1;
@@ -836,6 +837,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = "com.alisoftware.Dip-tvOSTests";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = appletvos;
+				SWIFT_VERSION = 3.0;
 				TVOS_DEPLOYMENT_TARGET = 9.0;
 			};
 			name = Debug;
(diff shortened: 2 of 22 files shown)


patrick-lind:
I kept your naming conventions in place since I noticed you changed them all back the first time around.. sorry about that! In the future you may consider giving in and conforming to Apple's recommendations for enum names and etc. to make conversions smoother :)

ilyapuchka:
Thanks @patrick-lind, I'll take a look later. 
The reason why I reverted some changes in api semantics in your previous pr is that I think we should not break api for users just to support Swift 3 or it's naming guidelines. That can be a separate major release. At most we should consider first deprecating old and providing new apis, not removing them completely.
I'm going to create an issue where I will describe how I see the release process in lights of Swift 3 release in September, but unfortunately I didn't have enough time to do that yet.

patrick-lind:
Not a problem! I'm sorry I did not even think to preserve your previous semantics. That said, converting to Swift 3 is a perfect time to break the api because... It's going to break regardless of what you do :) In this PR I've kept all of the api semantics the same so it's a fairly small subset of changes.. let me know if you have any questions when you get around to it!

ilyapuchka:
I'm a bit confused with this beta - what Swift release does it use?

Sources/Utils.swift
```diff
@@ -143,4 +143,5 @@ extension Collection where Index: Comparable, Self.Indices.Index == Index {
   
 #else
   import Foundation
+  typealias RecursiveLock = NSRecursiveLock
```
ilyapuchka: so "NS" is back again? 😫

patrick-lind: Yes I'm afraid so 😞. Unless I'm missing something?

This is using Swift 3 on Xcode Beta 4.

Sent from my iPhone

On Aug 3, 2016, at 5:03 PM, Ilya Puchka <notifications@github.com<mailto:notifications@github.com>> wrote:


In Sources/Utils.swift<https://github.com/AliSoftware/Dip/pull/106#discussion_r73420359>:

> @@ -143,4 +143,5 @@ extension Collection where Index: Comparable, Self.Indices.Index == Index {
>
>  #else
>    import Foundation
> +  typealias RecursiveLock = NSRecursiveLock


so "NS" is back again? 😫

ilyapuchka:
@patrick-lind can you please check why travis build fails? It said that it can not open playground, so probably it requires running swift converter on it again.

patrick-lind:
The Swift converter did not catch some of the issues in the Auto-Injection section of the Playground. That said when I was going through it I did see some runtime errors that I wasn't quite sure how to resolve but it should at least compile now.

patrick-lind:
Looks like that did it finally! Sorry about all the extra checkins... Xcode was making it hard to figure out what wasn't building properly. I would still double check your playgrounds and your spaceship example.. There were some runtime errors happening when I tried them out but I'm not sure if those were there previously or not.

ilyapuchka:
I think the runtime issue in playground is happening because of this bug https://bugs.swift.org/browse/SR-2144
In this case making `ServerClientImp` a subclass of `NSObject` should fix it.

ilyapuchka:
Thanks for your help @patrick-lind !

Regarding failures that you saw in playground I was able to solve two of them. They are caused by swift 3 issues, I liked them in #107 . I will update this issue with workarounds.
One crash on "Containers Collaboration" page is not reproducible in tests so I will assume it happens only in playgrounds.

PR #119, opened 2016-09-09
Fix compile error when building with Xcode8/Swift 3 GM

mark-urbanthings:
Apple Swift version 3.0 (swiftlang-800.0.46.2 clang-800.0.38)

The GM compiler introduces compile error this submission addresses.


ilyapuchka:
Thanks for the fix!

PR #146, opened 2017-03-30
Enable use of custom or no logging function

Pr0Ger:
This is a copy of Swinject feature: https://github.com/Swinject/Swinject/blob/master/Sources/Container.Logging.swift

It will allow user to redirect Dip logs to file and/or some custom logger if it's required


ilyapuchka:
Thanks for contribution!

PR #153, opened 2017-04-09
Autowiring should prioritize the definition with the matching tag

nmarkovic04:
Problem:

```
// Add interactor
container.register(.shared) { (entryDataAdapter: EntryDataAdapter, dateFormatter: DateFormatter) -> AbstractAddEntryInteractor in
   return AddEntryInteractor(entryDataAdapter: entryDataAdapter, dateFormatter: dateFormatter)
}.resolvingProperties({ (container, interactor) in
   interactor.presenter = try! container.resolve(tag: container.context.tag)
})

// Same module, different interactor; tag "editEntry"
container.register(tag: "editEntry", factory: { (selectedEntryID: Int, entryDataAdapter: EntryDataAdapter) -> AbstractAddEntryInteractor in
    return EditEntryInteractor(selectedEntryID: selectedEntryID, dataAdapter: entryDataAdapter) as AbstractAddEntryInteractor
}).resolvingProperties({ (container, interactor) in
   interactor.presenter = try! container.resolve(tag: container.context.tag)
})

// this is where interactor gets resolved
container.register(.shared) { () -> AbstractAddEntryPresenter in
   return AddEntryPresenter()
}.resolvingProperties({ (container, presenter) in
   presenter.interactor = try! container.resolve(tag: container.context.tag)
   presenter.view = try! container.resolve(tag: container.context.tag)
   presenter.routing = try! container.resolve(tag: container.context.tag)
})
```

Autowiring fails with DipError.ambiguousDefinitions while resolving graph with "editEntry" tag because there are multiple definitions with different arguments, specifically this line

```
// AutoWiring.swift, line 67
if definitions.count > 1 && definitions[0].key.typeOfArguments != definitions[1].key.typeOfArguments
``` 

even though only one definition matches the tag, so the instance should correctly get resolved.

### DIFF
--- a/Sources/AutoWiring.swift
+++ b/Sources/AutoWiring.swift
@@ -62,9 +62,15 @@ extension DependencyContainer {
     let maximumNumberOfArguments = definitions.first?.definition.numberOfArguments
     definitions = definitions.filter({ $0.definition.numberOfArguments == maximumNumberOfArguments })
     definitions = order(definitions: definitions, byTag: key.tag)
-
-    //when there are several definitions with the same number of arguments but different arguments types
-    if definitions.count > 1 && definitions[0].key.typeOfArguments != definitions[1].key.typeOfArguments {
+    
+    let firstDefinitionHasAMatchingTag = (key.tag == definitions[0].key.tag)
+    let secondDefinitionHasADifferentTag = definitions.count > 1 && (key.tag != definitions[1].key.tag)
+    
+      //when there are several definitions but only one matches the tag
+    if firstDefinitionHasAMatchingTag && secondDefinitionHasADifferentTag{
+      return definitions[0]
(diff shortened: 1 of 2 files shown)


ilyapuchka:
Hi @nmarkovic04! Thanks for your contribution!
Looks like there is an issue in this case indeed, thanks for pointing it out! While your fix solves it, I think it should be address in a different way.  Currently autowiring happens in one pass and it searches for definitions either matching tag or with `nil` tag (which we use as a fallback) and selects one with most number of arguments. When tag is used it looks like this strategy can lead to other cases when wrong definition is used.

Basically auto-wiring should happen in two passes, first should search for definitions for resolved type _strictly_ matching tag, second (if it's needed, when tag is not nil) should search for definitions for resolved type with `nil` tag. This way it will be more inline with a fallback strategy used in "normal" resolution flow.

So I will close this and will do PR for release/5.1 shortly
Thanks again for reporting!

PR #169, opened 2017-07-29
Fix: containers with common collaborator share resolved instances for their own definitions

### LINKED ISSUE #168: Question: How to create isolated containers
How can I create two isolated child containers whereby I have

a) A root container with registered singleton services
b) Two child containers which are independent of each other and may even have similar & conflicting registrations.

The children have access to the Root, but not each other. The Root doesn't have access to the children.

This is in the vain of [Daggers Subcomponents](https://google.github.io/dagger/subcomponents) and [Guices Private Module](https://google.github.io/guice/api-docs/latest/javadoc/index.html?com/google/inject/PrivateModule.html) or the always hideous [Swinject Container Hierarchy](https://github.com/Swinject/Swinject/blob/master/Documentation/ContainerHierarchy.md)

An example of this would be where I have two flows, one for logged in and one for unlogged in users.

```
class ServiceA {}

  class Password {
    let text: String
    let service : ServiceA

    init(text:String, service: ServiceA) {
      self.text = text
      self.service = service
    }
  }


  func testIsolationContainer() {

    let rootContainer = DependencyContainer()

    var count = 0
    rootContainer.register(.singleton) { () -> ServiceA in
      count = count + 1
      return ServiceA()
    }

    let loggedIn = DependencyContainer()
    loggedIn.register(.singleton) { Password(text: "1234", service:$0) }

    let unloggedIn = DependencyContainer()
    unloggedIn.register(.singleton) { Password(text: "-none-", service:$0) }

    // NOTE: Isolated containers
    loggedIn.collaborate(with: rootContainer) //Isolated containers
    unloggedIn.collaborate(with: rootContainer)//Isolated Container

    let passwordA = try! unloggedIn.resolve() as Password

    XCTAssert(passwordA.text == "-none-") //<< FAILS
    XCTAssert(count == 1) ////<< Works

    let passwordB = try! loggedIn.resolve() as Password
    XCTAssert(passwordB.text == "1234")  //<<<Works
    XCTAssert(count == 1)
  }
```

* I think the behavior changed with bug https://github.com/AliSoftware/Dip/issues/148 : Prior to this, the test failed differently; The containers were isolated but the SerivceA was created twice. 

* Having read the Collaboration code, both unlogged and logged in are bound together to be collaborators since it recurses through the dependencies->dependancies.

 * If I remove the ```loggedIn.collaborate(...)``` lines, it fails because it can't find the service

* The option to go directly to rootContainer.resolve() instead works...
```loggedIn.register(.singleton) { Password(text: "1234", service:rootContainer.resolve()) }```
but is not very desirable because it means when I setup my containers, I have to have extra knowledge about how the dep was registered and with what container, somewhat defeating the purpose of Dip.

**Thoughts?**

### PR DESCRIPTION
**Pull request by ilyapuchka**
Resolves #168

### DIFF
--- a/Sources/Dip.swift
+++ b/Sources/Dip.swift
@@ -167,12 +167,15 @@ extension DependencyContainer {
     /// The label of the property where resolved instance will be auto-injected.
     private(set) public var injectedInProperty: String?
     
+    let inCollaboration: Bool
+    
     var logErrors: Bool = true
     
-    init(key: DefinitionKey, injectedInType: Any.Type?, injectedInProperty: String?) {
+    init(key: DefinitionKey, injectedInType: Any.Type?, injectedInProperty: String?, inCollaboration: Bool) {
       self.key = key
       self.injectedInType = injectedInType
       self.injectedInProperty = injectedInProperty
+      self.inCollaboration = inCollaboration
     }
     
     public var debugDescription: String {
@@ -196,7 +199,7 @@ extension DependencyContainer {
 
   /// Pushes new context created with provided values and calls block. When block returns previous context is restored.
   /// When popped to initial (root) context will release all references to resolved instances and call `Resolvable` callbacks.
-  func inContext<T>(key aKey: DefinitionKey, injectedInType: Any.Type?, injectedInProperty: String? = nil, logErrors: Bool! = nil, block: () throws -> T) rethrows -> T {
+  func inContext<T>(key aKey: DefinitionKey, injectedInType: Any.Type?, injectedInProperty: String? = nil, inCollaboration: Bool = false, container: DependencyContainer? = nil, logErrors: Bool! = nil, block: () throws -> T) rethrows -> T {
     let key = aKey
     return try threadSafe {
       let currentContext = self.context
@@ -207,6 +210,10 @@ extension DependencyContainer {
         //clean instances pool if it is owned not by other container
         if context == nil {
           resolvedInstances.resolvedInstances.removeAll()
+          for (key, instance) in resolvedInstances.sharedWeakSingletons {
+            if resolvedInstances.sharedWeakSingletons[key] is WeakBoxType { continue }
+            resolvedInstances.sharedWeakSingletons[key] = WeakBox(instance)
+          }
           for (key, instance) in resolvedInstances.weakSingletons {
             if resolvedInstances.weakSingletons[key] is WeakBoxType { continue }
             resolvedInstances.weakSingletons[key] = WeakBox(instance)
@@ -222,7 +229,8 @@ extension DependencyContainer {
       context = Context(
         key: key,
         injectedInType: injectedInType,
-        injectedInProperty: injectedInProperty
+        injectedInProperty: injectedInProperty,
+        inCollaboration: inCollaboration
       )
       context.logErrors = logErrors ?? currentContext?.logErrors ?? true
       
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by jtwigg** [comment]
@ilyapuchka This fixes the original test where there's two isolated containers but the following test fails.


a) There is a rootContainer
b) There are two isolated child containers, each collaborates with root.
   i ) not logged in has a valid registry of Password
  ii) logged in has **NO** valid registry

When resolving password on both containers, the first one should pass and the second resolution should fail 
**BUT**

**The two child containers are leaking into each other. The second container has access to the first containers registry.**

```

  class ServiceA {}

  class Password {
    let text: String
    let service : ServiceA

    init(text:String, service: ServiceA) {
      self.text = text
      self.service = service
    }
  }


  func testIsolationContainer2() {

    let rootContainer = DependencyContainer()

    var count = 0
    rootContainer.register(.singleton) { () -> ServiceA in
      count = count + 1
      return ServiceA()
    }

  //NOTE: Nothing is registered
    let loggedIn = DependencyContainer()

    let unloggedIn = DependencyContainer()
    unloggedIn.register() { Password(text: "-none-", service:$0) }

    // NOTE: Isolated containers
    loggedIn.collaborate(with: rootContainer) //Isolated containers
    unloggedIn.collaborate(with: rootContainer)//Isolated Container

    let passwordA = try! unloggedIn.resolve() as Password

    XCTAssert(passwordA.text == "-none-")
    XCTAssert(count == 1) ////<< Works

    let passwordB : Password? = try? loggedIn.resolve() as Password

    XCTAssertNil(passwordB) //<<<< FAILS. Should be nil,  but its "-none-"
    XCTAssert(count == 1)
  }

```

**Comment by jtwigg** [comment]
@ilyapuchka Here's a second failure in the tests you applied

** Root Containers should not have access to their children's registry **

I added a check in the test you created: 

```
  func testThatContainersShareTheirSingletonsOnlyWithCollaborators() {
    let container = DependencyContainer()
    container.register(.singleton) { RootService() }
    
    let collaborator1 = DependencyContainer()
    collaborator1.register(.singleton) {
      ServiceClient(name: "1", service: $0)
    }
    
    let collaborator2 = DependencyContainer()
    collaborator2.register(.singleton) {
      ServiceClient(name: "2", service: $0)
    }


    collaborator1.collaborate(with: container)
    collaborator2.collaborate(with: container)

    //Root client should not have access to its childrens services
    let rootClient = try? container.resolve() as ServiceClient
    XCTAssertNil(rootClient) //<<<<<< FAILS

    
    let client2 = try! collaborator2.resolve() as ServiceClient
    let client1 = try! collaborator1.resolve() as ServiceClient
    
    XCTAssertEqual(client1.name, "1")
    XCTAssertEqual(client2.name, "2")
    XCTAssertTrue(client1.service === client2.service)
  }
```

I believe this is do to the change you made where adding a collaborator to yourself ALSO forces that to collaborate with you. 

```
public func collaborate(with containers: [DependencyContainer]) {
    _collaborators += containers
    for container in containers {
      container._collaborators += [self] //<<<<<<<<<<< Might not be correct
      container.resolvedInstances.sharedSingletonsBox = self.resolvedInstances.sharedSingletonsBox
      container.resolvedInstances.sharedWeakSingletonsBox = self.resolvedInstances.sharedWeakSingletonsBox
      updateCollaborationReferences(between: container, and: self)
    }
  }
```

ilyapuchka:
Thanks for adding those test cases, I'll take a look later.
Regarding root container having no access to others - it is by design. It's not parent - child relationship, collaboration is bidirectional.

jtwigg:
I agree, you did make it bidirectional in this commit, e7d8fb41e1c8a852c603d7227ddc5f4f2bc1f816  but I think that means we'll never have isolated containers.

So I can see now that collaborations has a completely different use case. The wiki talks about Modules (good) but its clear now you didn't include isolation as a property of a module, since its bi directional. That is something I was expecting and rather counting on.

**My question stands: https://github.com/AliSoftware/Dip/issues/168 How do you intend people to create isolated hierarchical containers? Or is that simply not supported?**

ilyapuchka:
My vision on that is that you can not have everything. If your container collaborate their are connected into a bidirectional graph, and library goes around this graph to resolve instance. In in this graph there are several ambiguous definitions and there is a rout to both of them - there is no way to say what variant will be used (it depends on the order of containers references).
In your case I would suggest either to use named definitions (for that you will need to store tag that represents state of your app, like loggedIn and notLoggedIn) and use tag to disambiguate your requests at runtime, or do not keep both containers around at the same time, but this will make you to recreate containers graph. We can also have a method to stop collaboration which can make this task easier.

PR #176, opened 2017-09-16
Swift 4 migration

ilyapuchka:


DenHeadless:
Do you have plans on where to release those fixes? Seems like CI jobs are completed successfully, only pod lint validation failed for some reason.

Issue #185, opened 2018-02-27
Nil in AutoInject properties in every VC that not in first uiContainer


abesmon:
Code to reproduce

```
        class BaseVC: UIViewController {}

        class FooVC: BaseVC {}
        class FooOneVC: FooVC, StoryboardInstantiatable {
            private(set) var nest = Injected<Nest>()
            class Nest {}
        }
        class FooSecVC: FooOneVC {}

        class BarVC: BaseVC {}
        class BarOneVC: BarVC {
            private(set) var nest = Injected<Nest>()
            class Nest {}
        }
        class BarSecVC: BarOneVC {}

        // MARK: foo part
        let fooUICont = DependencyContainer { container in
            container.register(.shared) { FooOneVC() }
            container.register(.shared) { FooSecVC() }
        }

        let fooDataCont = DependencyContainer { container in
            container.register(.shared) { (controller: FooOneVC) -> FooOneVC.Nest in
                return FooOneVC.Nest()
            }
        }

        fooUICont.collaborate(with: fooDataCont)
        fooDataCont.collaborate(with: fooUICont)

        try! fooUICont.bootstrap()
        try! fooDataCont.bootstrap()
        DependencyContainer.uiContainers.append(fooUICont)


        // MARK: bar part
        let barUICont = DependencyContainer { container in
            container.register(.shared) { BarOneVC() }
            container.register(.shared) { BarSecVC() }
        }

        let barDataCont = DependencyContainer { container in
            container.register(.shared) { (controller: BarOneVC) -> BarOneVC.Nest in
                return BarOneVC.Nest()
            }
        }

        barUICont.collaborate(with: barDataCont)
        barDataCont.collaborate(with: barUICont)

        try! barUICont.bootstrap()
        try! barDataCont.bootstrap()
        DependencyContainer.uiContainers.append(barUICont)

        let fooSecVC = FooSecVC()
        fooSecVC.setValue(nil, forKey: "dipTag")
        print(fooSecVC.nest.value)

        let barSecVC = BarSecVC()
        barSecVC.setValue(nil, forKey: "dipTag")
        print(barSecVC.nest.value)
```

**What was expected:**
`fooSecVC.nest.value` and `barSecVC.nest.value` would be not nil

**What happens:**
`fooSecVC.nest.value` not nil, but `barSecVC.nest.value` is nil.
And as seen in our main application, all VC's that are not in first uiContainer, works wrong with `Injected<T>()`

abesmon:
[...]

PR #186, opened 2018-02-27
autoinject fails

abesmon:
refers to #185 

i don't know if it's the right solution and if this one will not lead to another issues, but this one works.

Seems like these files was changed while parent system implementation, so it needs some review


ilyapuchka:
This will affect parent-child relationships, probably breaking it

ilyapuchka:
revert #189 should address this

PR #189, opened 2018-04-26
Revert #171

ilyapuchka:
This pr reverts #171 until further investigation as it seems that it introduced regression.


ilyapuchka:
cc @john-twigg-ck

john-twigg-ck:
@ilyapuchka Can you describe the regression? 
I can show you all the further updates we've made to the library since then.

ilyapuchka:
It is #185 and there is related issue in DipUI. If you can, just make a PR with those updates.

PR #191, opened 2018-05-25
Removed extension of ImplicitlyUnwrappedOptional

Gruppio:
ImplicitlyUnwrappedOptional was already deprecated and it has been removed in swift 4.2
If you don't remove this extension Dip won't compile with Swift 4.2

### DIFF
--- a/Sources/Utils.swift
+++ b/Sources/Utils.swift
@@ -48,12 +48,6 @@ extension Optional: BoxType {
   }
 }
 
-extension ImplicitlyUnwrappedOptional: BoxType {
-  var unboxed: Any? {
-    return self ?? nil
-  }
-}
-
 class Box<T> {
   var unboxed: T
   init(_ value: T) {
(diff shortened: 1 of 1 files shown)


ilyapuchka:
@Gruppio thanks, but this will be merged only when Swift 4.2 is released

beefon:
Is there a way to compile this conditionally?

beefon:
https://github.com/apple/swift-evolution/blob/master/proposals/0020-if-swift-version.md

beefon:
I assume this PR might be updated with 
```
#if swift(<4.2)
extension ImplicitlyUnwrappedOptional: BoxType {		
   var unboxed: Any? {		
     return self ?? nil		
   }		
 }
#endif

```

Gruppio:
I've add the check if swift is >= 4.1, so it is removed a warning in the current stable release ( 4.1 )

PR #194, opened 2018-06-06
Adjust Dip scheme to remove code coverage for Carthage Static Frameworks

kaosdg:
... and add a new Testing scheme that has code coverage.

We're currently using Carthage static frameworks (https://github.com/Carthage/Carthage/blob/master/Documentation/StaticFrameworks.md) in our applications to increase app load times.

However, when Carthage builds Dip for static frameworks, the code coverage option on the Dip scheme leads to the following errors:

```
Undefined symbols for architecture x86_64:
  "___llvm_profile_runtime", referenced from:
```

By disabling code coverage on the scheme that Carthage builds, this issue is resolved.

I've added a separate Scheme solely for tests that enables code coverage, and adjusted the .travis.yml settings to use that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ matrix:
     - os: linux
   include:
     - script:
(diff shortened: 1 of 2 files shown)


ilyapuchka:
It seems to me as this is entirely Carthage issue or something that is not officially supported, but test coverage is something that can be sacrificed, so no need to extra scheme, it can be just turned off. Done in 33993df68a332c1748e8e16d5d977eb69eb879c0

PR #201, opened 2018-08-28
Fix for Swift 4.2 changes in Optional<->Any conversions

stevenkramer:
In Xcode 10 beta 5, the unboxing of optional values started to fail, because the code simultaneously tried to test for an empty optional and a successful cast. Breaking this up in two steps fixes the issue.

### DIFF
--- a/Sources/Resolve.swift
+++ b/Sources/Resolve.swift
@@ -199,7 +199,7 @@ extension DependencyContainer {
      That happens because when Optional is casted to Any Swift can not implicitly unwrap it with as operator.
      As a workaround we detect boxing here and unwrap it so that we return not a box, but wrapped instance.
      */
-    if let box = resolvedInstance as? BoxType, let unboxed = box.unboxed as? T {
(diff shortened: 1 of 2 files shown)


ilyapuchka:
have you tried swift42 branch?

stevenkramer:
Yes, was using that since `dadb688` on previous betas. This started to manifest itself around beta 4.

stevenkramer:
Happened trying to auto-wire an optional initializer argument.

ilyapuchka:
ah, I didn't notice that you opened this for this branch, my bad

ilyapuchka:
@stevenkramer can you add a test that is failing without this change?

stevenkramer:
@ilyapuchka done

ilyapuchka:
This is not related to Swift 4.2, it also happens with Swift 4. It's due to reusing previously resolved instances - if the order of resolves is different in your test it works, but it can't reuse instances resolved for optional type. It also only happens for structs, not classes.

PR #221, opened 2019-04-02
Print nested types in errors, hash keys using full typenames

stevenkramer:
Hi Ilya, we're using nested types a lot and the "leafs" have identical names, which leads to conflicts when type matching and non-specific log messages. This change uses the fully scoped names.


Sources/DipError.swift
```diff
@@ -77,18 +77,21 @@ public enum DipError: Error, CustomStringConvertible {
...
-      return "Resolved instance \(resolved ?? "nil") does not implement expected type \(key.type)."
+      return "Resolved instance \(resolved ?? "nil") does not implement expected type \(reflect(key.type))."
     }
   }
-  
 }
 
+func reflect(_ type: Any.Type) -> String
+{
+  return String(reflecting: type)
```
ilyapuchka: can you please just use this rather than introducing free function?

Sources/Definition.swift
```diff
@@ -41,7 +41,7 @@ public struct DefinitionKey: Hashable, CustomStringConvertible {
   }
   
   public var description: String {
-    return "type: \(type), arguments: \(typeOfArguments), tag: \(tag.desc)"
+    return "type: \(reflect(type)), arguments: \(typeOfArguments), tag: \(tag.desc)"
```
ilyapuchka:
```suggestion
    return "type: \(String(reflecting: type)), arguments: \(typeOfArguments), tag: \(tag.desc)"
```

Sources/DipError.swift
```diff
@@ -77,18 +77,21 @@ public enum DipError: Error, CustomStringConvertible {
   public var description: String {
     switch self {
     case let .definitionNotFound(key):
-      return "No definition registered for \(key).\nCheck the tag, type you try to resolve, number, order and types of runtime arguments passed to `resolve()` and match them with registered factories for type \(key.type)."
+      return "No definition registered for \(key).\nCheck the tag, type you try to resolve, number, order and types of runtime arguments passed to `resolve()` and match them with registered factories for type \(reflect(key.type))."
```
ilyapuchka:
```suggestion
      return "No definition registered for \(key).\nCheck the tag, type you try to resolve, number, order and types of runtime arguments passed to `resolve()` and match them with registered factories for type \(String(reflecting: key.type))."
```

Sources/DipError.swift
```diff
@@ -77,18 +77,21 @@ public enum DipError: Error, CustomStringConvertible {
...
     case let .autoInjectionFailed(label, type, error):
-      return "Failed to auto-inject property \"\(label.desc)\" of type \(type). \(error)"
+      return "Failed to auto-inject property \"\(label.desc)\" of type \(reflect(type)). \(error)"
```
ilyapuchka:
```suggestion
      return "Failed to auto-inject property \"\(label.desc)\" of type \(String(reflecting: type)). \(error)"
```

Sources/DipError.swift
```diff
@@ -77,18 +77,21 @@ public enum DipError: Error, CustomStringConvertible {
...
     case let .autoWiringFailed(type, error):
-      return "Failed to auto-wire type \"\(type)\". \(error)"
+      return "Failed to auto-wire type \"\(reflect(type))\". \(error)"
```
ilyapuchka:
```suggestion
      return "Failed to auto-wire type \"\(String(reflecting: type))\". \(error)"
```

Sources/DipError.swift
```diff
@@ -77,18 +77,21 @@ public enum DipError: Error, CustomStringConvertible {
...
     case let .ambiguousDefinitions(type, definitions):
-      return "Ambiguous definitions for \(type):\n" +
+      return "Ambiguous definitions for \(reflect(type)):\n" +
```
ilyapuchka:
```suggestion
      return "Ambiguous definitions for \(String(reflecting: type)):\n" +
```

Sources/DipError.swift
```diff
@@ -77,18 +77,21 @@ public enum DipError: Error, CustomStringConvertible {
...
         definitions.map({ "\($0)" }).joined(separator: ";\n")
     case let .invalidType(resolved, key):
-      return "Resolved instance \(resolved ?? "nil") does not implement expected type \(key.type)."
+      return "Resolved instance \(resolved ?? "nil") does not implement expected type \(reflect(key.type))."
```
ilyapuchka:
```suggestion
      return "Resolved instance \(resolved ?? "nil") does not implement expected type \(String(reflecting: key.type))."
```

PR #225, opened 2019-06-30
Property wrappers

ilyapuchka:
By adding `@propertyDelegate` to our existing auto-injection wrappers instead of

```swift
let client = Injected<Client>()
```

in Swift 5.1 we can write

```swift
@Injected var client: Client? = nil
```

~Note: as of Xcode 11 beta 1 `()` are needed after property delegate name~
~UPD: Not an issue any more with refactoring of property wrappers into structs, but still requires initial `nil` value provided explicitly when using property wrapper (example updated) in class (not needed in structs)~

UPD: `= nil` is not needed any more


dchohfi:
hey @ilyapuchka, thoughts on this? I might be able to pick it and see what is left

PR #229, opened 2019-10-21
Bump watchos deployment target to 3.0

dchohfi:
Apple now requires that watchOS deployment target are minimum locked at 3.0 version.

https://forums.developer.apple.com/thread/108853

Could you create a tag after merging so we can use it on Carthage?

Thank you!


dchohfi:
hey @AliSoftware, any thoughts on this? I'm blocked to submit a watch target using Carthage. thank you.

AliSoftware:
I'll let @ilyapuchka — who's now the main maintainer — to double-review and merge this

